Issue #32, opened 2026-09-03
Refactor simulation ownership for deterministic state, entity lifecycle, and scalable physics


xtreemze:
Parent: #29
Related: #30, #31

## Goal

Make the game simulation understandable and testable without erasing the emergent physics that defines Defend.

The current code stores most mutable state in global objects and couples gameplay rules directly to Babylon mesh/impostor lifecycle callbacks. This makes it difficult to distinguish game state, rendering state, and physics-object ownership, and increases the chance that modernization changes gameplay accidentally.

## Target direction

Move incrementally toward explicit simulation domains:

- economy/energy;
- wave progression;
- enemy state and finite lifetime;
- tower placement, level, cost, aging/degradation;
- projectile damage/recovery semantics;
- spatial/physics binding;
- presentation/effects/audio as observers of simulation outcomes where feasible.

This does **not** require replacing the physics simulation with a deterministic grid simulation. Physics remains authoritative for collision, momentum, knockback, obstruction and ejection.

## Invariants

- One authoritative owner for each entity lifecycle.
- No double charge/refund/destroy from overlapping Babylon callbacks and timers.
- Timers and render callbacks are cancellable and removed when entities die/dispose.
- Pure calculations should not require Babylon globals.
- Stochastic enemy decisions should accept an injectable RNG/seed where practical.
- Entity IDs/types/levels should be explicit rather than parsed from mesh-name character positions.
- Simulation data should not depend on DOM elements.
- Effects/audio should consume events/state changes rather than become hidden gameplay authority.
- Keep hot per-frame paths allocation-conscious and suitable for mobile hardware.

## Known suspicious areas worth resolving while refactoring

- `shotClearsTower()` computes a raycast result but currently returns `true` unconditionally and its call site is disabled.
- Tower upgrade logic parses a level from a mesh-name character and level 3 can be rebuilt as level 3.
- Tower aging/degradation mixes disposal, refunds and reconstruction and needs certification against #30.
- Wave generation and callback unregistering rely on closures/functions that may not be removable by identity.
- Global arrays (`allEnemies`, `allTowers`, impostors/positions) can drift from actual scene ownership.
- Enemy placement collision checks appear logically inconsistent and should be verified rather than silently preserved.

## Parallelization

This issue should be implemented through small PRs with explicit ownership boundaries, for example:

1. Introduce typed gameplay constants/calculation helpers with no behavior change.
2. Introduce explicit entity metadata/IDs and stop parsing mesh names.
[...]

Issue #30, opened 2026-09-03
Baseline certification: capture deterministic gameplay invariants before modernization


xtreemze:
Parent: #29

## Goal

Create a reliable behavioral baseline so engine/toolchain/refactor PRs can prove that they preserved Defend's essential mechanics.

This work is intentionally split between GitHub implementation and the local Codex executor. The local executor has the checkout/runtime and should perform browser execution, historical dependency installation where needed, profiling, screenshots, and interactive physics checks; findings should be posted here and attached to implementation PRs.

## Contract to measure

At minimum certify:

- energy ramps to the intended starting balance;
- placing each tower level charges the intended cost exactly once;
- projectile impacts reduce enemy HP and replenish energy using the intended recovery ratio, capped at max balance;
- an enemy reaching the central bank damages energy based on its remaining threat/HP;
- enemy HP decays over decision ticks even without projectile damage;
- projectile impulse produces real knockback and falling below the arena removes enemies;
- level 2/3 towers differ in range/rate/projectile mass/damage in the intended direction;
- tower lifetime causes degradation and eventual removal, without accidental double charging/refunding;
- wave advancement, resume/current-wave persistence, victory, defeat, and restart remain coherent;
- placement grid and occupied central cells cannot overlap;
- direct tap placement/upgrade remains usable with pointer/touch input.

## Desired implementation

Prefer a layered harness rather than one fragile end-to-end test:

1. Extract/test pure calculations where possible (cost, damage, recovery, wave composition, decay, timing relationships).
2. Add deterministic seeded simulation seams for stochastic choices where practical.
3. Add a small browser smoke/certification flow for Babylon/physics behavior that cannot be represented faithfully in unit tests.
4. Record a compact baseline fixture/report suitable for comparing modernization PRs.

Do not freeze obvious bugs as requirements. If observed runtime behavior conflicts with #29, document the discrepancy here before encoding it.

## Local Codex tasks

- establish whether the historical project builds/runs as-is on a current machine;
- record exact install/build/runtime obstacles;
- capture a baseline browser session and current console errors;
- verify current tower degradation, collision economy, knockback/ejection, and wave lifecycle;
- report any mismatch between README/help text and actual behavior;
- once a harness PR exists, run it against both baseline and proposed modernization branches.

## Completion

This issue is complete when subsequent PRs have a repeatable command/checklist that can distinguish a preserved gameplay loop from an accidental rewrite.

xtreemze:
[...]

PR #41, opened 2026-09-03
refactor: stop parsing tower level from mesh names

xtreemze:
Refs #32
Related #30

## Ownership boundary

This PR changes only tower identity metadata and the upgrade handler's level lookup. It does not alter tower costs, upgrade tier transitions, lifetimes, firing behavior, physics, or the known level-3 refresh semantics.

## Changes

- attach explicit `defendTower` metadata (`level`, `index`) to each tower base when it is constructed;
- make upgrade handling read the explicit level instead of `pickedMesh.name[10]`;
- refuse the upgrade path when a tagged tower base lacks valid level metadata rather than disposing an object based on malformed name parsing.

## Why

Issue #32 identifies mesh-name parsing as a brittle lifecycle coupling. Tower names should remain useful for diagnostics, but gameplay state should not depend on a character offset in a string. This creates the first explicit entity metadata seam without changing the economy/lifecycle decisions still awaiting #30 certification.

## Validation

Online review confirms the change is limited to `Tower.ts` and `upgradeTower.ts`; the existing level switch and affordability logic are intentionally untouched.

Local Codex validation requested:

- [ ] historical TypeScript/BabylonJS stack compiles the `Mesh.metadata` access;
- [ ] placing L1/L2/L3 towers attaches correct metadata;
- [ ] tapping L1 upgrades to L2 with the same charge as baseline;
- [ ] tapping L2 upgrades to L3 with the same charge as baseline;
- [ ] existing level-3 behavior remains unchanged for this PR;
- [ ] degradation/reconstruction produces new tower bases with correct metadata;
- [ ] no change to tower placement, collision, or targeting behavior.

Do not fold affordability-boundary or level-3 gameplay decisions into this PR; those should remain separate after #30 evidence.

### DIFF
--- a/src/js/tower/Tower.ts
+++ b/src/js/tower/Tower.ts
@@ -13,6 +13,29 @@ import { towerGlobals, projectileGlobals } from "../main/globalVariables";
 import { Position2D } from "../enemy/Enemy";
 import { EnemySphere } from "../enemy/enemyBorn";
 
+interface TowerMetadata {
+	level: number;
+	index: number;
+}
+
+function setTowerMetadata(tower: Mesh, level: number, index: number): void {
+	const metadata = tower.metadata || {};
+	metadata.defendTower = { level, index } as TowerMetadata;
+	tower.metadata = metadata;
+}
+
+function towerLevel(tower: Mesh): number {
+	const metadata = tower.metadata;
+	if (
+		metadata &&
+		metadata.defendTower &&
+		typeof metadata.defendTower.level === "number"
+	) {
+		return metadata.defendTower.level;
+	}
+	return 0;
+}
+
 class Tower {
 	constructor(
 		level: number = 1 | 2 | 3,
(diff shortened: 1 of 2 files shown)


xtreemze:
Merge review after concurrent #40/#42: current comparison against `master` still isolates this PR to tower metadata and upgrade-level lookup; the #40 helper fix is already present on the base and does not need replaying. Local Codex: certify head `981f16cc7b9082cf53a690d802bdb3bb7dcb7822` using the checklist in the PR body, with particular attention to BabylonJS 3.x `Mesh.metadata`, degradation/reconstruction metadata, and unchanged L1→L2/L2→L3/level-3 tap behavior. Post commands and exit/runtime results before merge.

xtreemze:
Superseded by #44. `master` advanced concurrently in `Tower.ts` via #40; rather than force-rebase this branch or risk reverting another executor's fix, I recreated the metadata change from current `master` and reopened it cleanly with zero-behind comparison.

PR #36, opened 2026-09-03
Extract pure gameplay economy calculations

xtreemze:
## Scope

Linked issues: #30, #32

Ownership boundary: projectile-hit energy recovery and central-bank collision damage calculations only.

## Change

- add `src/js/gameplay/economy.ts`, a Babylon/Cannon-free module for the two core energy formulas;
- route projectile-hit recovery through the pure calculation while preserving the existing `projectileHitPoints * energyRecoveryRatio` behavior and max-balance cap;
- route central-bank collision damage through the pure calculation while preserving `enemyHitPoints / 2` and intentionally allowing the balance to cross below zero for the existing defeat path.

This is the first deterministic seam for the #30 characterization effort. It does not change collision registration, enemy lifecycle, projectile physics, tower behavior, wave timing, rendering, or sound.

## Gameplay impact

- [x] No intended gameplay/balance change
- [ ] Intended gameplay change is documented in the linked issue

Preserved #29 invariants:

- offense still sustains defense at exactly the existing recovery ratio;
- remaining enemy threat/HP still determines damage to the defended energy bank;
- physics collisions remain authoritative for deciding when either transition occurs.

## Validation

Performed online:

- branch is based on current `master` after #34;
- reviewed a 3-file diff: 37-line pure module plus two narrow call-site substitutions;
- formulas and cap/negative-balance semantics were compared directly with the legacy expressions.

Not validated online / requires local Codex:

- [x] build/install
- [x] browser interaction
- [x] physics behavior
- [ ] performance/profile
- [ ] touch/input
- [ ] audio
- [ ] PWA/offline
- [ ] other

Local certification should confirm that a projectile collision changes enemy HP and energy exactly as before, a bank collision damages energy based on remaining HP, and the historical project still compiles with the new module/imports.

## Concurrency

No overlap intended with toolchain work in #31 or presentation work in #33. Simulation refactors under #32 may build on this pure module rather than duplicating the formulas.

This PR is intentionally a draft until the local Codex executor posts build/runtime certification.

PR #35, opened 2026-09-03
chore: make tooling commands safe by default

xtreemze:
Refs #31

## Ownership boundary

This PR changes only repository command semantics in `package.json`. It does not change gameplay, rendering, physics, dependencies, or generated output.

## Changes

- make `npm run lint` non-mutating;
- preserve explicit mutation as `npm run lint:fix`;
- make `npm run build` a one-shot production build suitable for validation/automation;
- preserve watch behavior as `npm run build:watch`;
- fix the invalid `--hots` flag in `soundTest` to `--hot`.

## Why

Concurrent AI executors and local certification should be able to run ordinary validation without unexpectedly rewriting source files or entering an indefinite watch process. This is the first low-risk slice of #31's reproducible/non-invasive tooling stage.

## Validation

Online inspection confirms this is a single-file command-only change. No runtime or gameplay behavior is modified.

Local Codex should verify on the historical dependency graph:

- [ ] `npm run lint` terminates without modifying tracked files;
- [ ] `npm run lint:fix` remains available when mutation is intentional;
- [ ] `npm run build` terminates after one production build;
- [ ] `npm run build:watch` retains the previous watch behavior;
- [ ] `npm run soundTest` starts webpack-dev-server without rejecting the CLI flag.

Do not block unrelated modernization work on this PR; its ownership is limited to script semantics.

PR #43, opened 2026-09-03
Use explicit metadata for tower upgrade levels

xtreemze:
Refs #32, #30

## Ownership boundary

Tower-level identity used by the pointer upgrade path only.

## Change

- attach explicit `{ level }` metadata to each tower base mesh at construction;
- read the level only after the pointer target has been verified as a `towerBase`;
- remove the brittle `parseInt(pickedMesh.name[10])` dependency.

Tower names are intentionally left unchanged for diagnostics/backward familiarity. Upgrade costs and the existing level 1→2, 2→3, and current level-3 behavior are not changed in this PR.

## Why

#32 calls out mesh-name character parsing as hidden gameplay state. Tower level is simulation data and should be explicit rather than encoded in a presentation/debug string. This is a narrow prerequisite for later lifecycle and upgrade remediation.

## Gameplay impact

- [x] No intended balance change.
- [x] Existing upgrade branching and economy expressions are preserved.
- [x] No physics, rendering, audio, wave, or dependency changes.

## Online verification

- based on `master` after #40 (`38ff29a560f7453efb8e00b4790a1dbfc32e03a9`);
- two files changed: `Tower.ts` and `upgradeTower.ts`;
- 12 additions / 5 deletions;
- metadata is assigned before the mesh receives the `towerBase` tag;
- the accessor is invoked only after the tag/pick checks, so arbitrary picked meshes no longer have their names parsed.

## Local Codex certification

Please certify this head before merge:

1. TypeScript/build succeeds or report the exact historical-toolchain blocker.
2. Place a level-1 tower and upgrade 1→2 and 2→3; confirm the same costs are charged exactly once.
3. Verify level metadata survives the live Babylon mesh lifecycle through placement and each reconstruction.
4. Record the current behavior of tapping/upgrading an already-level-3 tower without changing it here; #32 tracks whether that behavior is a defect.
5. Confirm non-tower pointer taps remain harmless.

Post commands, exit codes, and observed results to this PR. Do not broaden this PR into the level-3 behavior fix.

## Concurrency

Independent from #35 tooling and #36 economy work. Any intentional change to upgrade caps/level-3 behavior should be a separate PR after baseline evidence.

### DIFF
--- a/src/js/tower/Tower.ts
+++ b/src/js/tower/Tower.ts
@@ -13,6 +13,10 @@ import { towerGlobals, projectileGlobals } from "../main/globalVariables";
 import { Position2D } from "../enemy/Enemy";
 import { EnemySphere } from "../enemy/enemyBorn";
 
+interface TowerMetadata {
+	level: number;
+}
+
 class Tower {
 	constructor(
 		level: number = 1 | 2 | 3,
@@ -25,6 +29,7 @@ class Tower {
 		let tower = towerGlobals.towerBaseMesh.clone(
 			name, undefined, true, false
 		) as Mesh;
+		tower.metadata = { level } as TowerMetadata;
 
 		tower.physicsImpostor = new PhysicsImpostor(
 			tower,
@@ -39,6 +44,10 @@ class Tower {
 		addTower(tower, level);
 		towerGlobals.allPositions = towerBasePositions(scene);
 	}
+}
+
+function towerLevel(tower: Mesh): number {
+	return (tower.metadata as TowerMetadata).level;
 }
 
 function towerBasePositions(scene: Scene) {
@@ -146,6 +155,7 @@ export {
 	Tower,
 	destroyTower,
 	towerBasePositions,
+	towerLevel,
 	shotClearsTower,
 	rotateTurret
 };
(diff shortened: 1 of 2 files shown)


xtreemze:
Reviewed against the overlapping metadata implementation I was preparing. This PR is the better ownership unit: smaller diff, same intended behavior preservation, and it already incorporates the current `master` tower-occlusion fix. I closed the duplicate #44 in favor of this one.

One deliberate boundary to keep: `towerLevel()` assumes tagged tower bases carry metadata. That is acceptable for the current constructor-owned lifecycle and keeps this PR narrow; if entity creation becomes more heterogeneous later, we can harden the accessor as part of the broader explicit-ID/lifecycle work rather than expanding this PR now.

xtreemze:
Local-certification routing update: this exact tower-metadata implementation is now included unchanged in aggregate draft #95 (`440b09ca0630cb1e25686e4b60d06bf4364c7af8`) on current master. Prefer certifying #95 once rather than rebuilding this branch separately. Keep this PR open as provenance; if the aggregate passes, merge #95 and close this PR as superseded.

Issue #53, opened 2026-09-03
Build a physics- and material-driven procedural audio engine


xtreemze:
Parent: #29
Related: #30, #31, #32, #33, #39, #45, #46, #52

## Goal

Replace the current event-name-driven sound layer with a procedural acoustic model whose output is derived from the actual simulation: contact impulse, relative velocity, mass, angular velocity, contact point, object/material pair, damage state, body scale, and listener transform.

The result should preserve Defend's generated/spatial sonic identity while making the world audibly communicate its physics.

## Current state

`src/js/main/sound.ts` already procedurally generates positional sounds through WAFXR/Tone. This is a good conceptual foundation, but most synthesis parameters are derived from event type, level, HP, or object counts rather than the actual collision.

Examples:
- turret shots derive frequency mainly from tower level;
- projectile damage derives frequency from enemy remaining HP;
- bank damage is a fixed sine/repeat shape;
- explosions derive frequency from enemy level;
- simultaneous-sound limits are global counters released by timers;
- listener position/orientation is updated every render from the active camera.

Collision callbacks in `hitEffect.ts`, `destroyOnCollide.ts`, and `currencyCollide.ts` currently discard most physically useful contact information.

## Acoustic event contract

Introduce a renderer/audio-engine-independent event produced by simulation/physics, conceptually:

```text
AcousticContact {
  event_id,
  time,
  kind,                 // impact, scrape, roll, fracture, launch, collapse, energy-transfer...
  position,
  normal,
  relative_velocity,
  normal_impulse,
  tangential_speed,
  angular_speed,
  effective_mass,
  body_a_size,
  body_b_size,
  material_a,
  material_b,
  damage_energy,
  seed
}
```

The exact fields depend on what Cannon/Rapier can expose reliably. Do not fabricate impulse values when the engine does not provide them; derive and document approximations from pre/post velocity and effective mass where necessary.

## Acoustic material model

Give gameplay materials explicit semantic acoustic properties independent of visual shaders. Suggested parameters:

- density / effective mass contribution;
- stiffness / hardness;
- damping / decay;
- roughness;
- resonant modes or mode ratios;
- excitation noise amount;
- brightness / high-frequency loss;
- friction/scrape character;
- nonlinear saturation amount;
- optional procedural variation range.

Initial semantic families might include:
- tower structure: dense metallic/ceramic body;
- projectile: hard compact metallic/energy body;
- enemy sphere: synthetic resonant/energy shell whose state changes as HP decays;
[...]

Issue #52, opened 2026-09-03
Evaluate a Rust/WASM simulation core and Bevy migration path


xtreemze:
Parent: #29
Related: #30, #31, #32, #45, #46

## Goal

Evaluate Rust/WASM as a first-class implementation path for Defend where it materially improves determinism, physics portability, shared native/web code, and long-term maintainability. Do not assume a full rewrite is automatically superior; compare a Rust simulation core behind the existing Babylon renderer against a full Bevy-based runtime.

## Why this is credible in 2026

Current upstream capabilities make this materially different from the 2019 environment:

- Bevy 0.19 is current and provides browser examples compiled to WASM for both WebGL2 and WebGPU;
- Bevy's web examples cover 3D rendering, input/picking, generated assets, spatial audio, fixed-timestep movement, persistence, and stress testing;
- Rapier integrates with Bevy, supports `wasm-bindgen`, serialization, and an `enhanced-determinism` mode intended to provide cross-platform determinism across supported IEEE-754 32/64-bit targets including WASM;
- Defend's gameplay is unusually well-suited to a deterministic data-oriented core: economy, waves, temporary towers, finite-lived enemies, projectile impulses, occupancy, collision-derived damage, and acoustic events are all separable from presentation.

Upstream references checked 2026-09-03:
- https://bevy.org/news/bevy-0-19/
- https://bevy.org/examples/
- https://bevy.org/examples-webgpu/
- https://rapier.rs/docs/user_guides/templates/getting_started_bevy/

### A. Rust/WASM simulation core + existing web renderer

Rust owns:
- entity IDs and lifecycle;
- fixed-step simulation clock;
- economy and wave state;
- deterministic random source;
- tower/enemy/projectile tier calculations;
- collision/contact semantic events once a Rust physics engine is adopted;
- serialization/replay/checkpoint state;
- physics-to-audio event descriptors.

TypeScript/Babylon initially owns:
- rendering;
- browser input/UI;
- current Cannon physics during the first bridge stage;
- Web Audio presentation.

This option lets the new pure gameplay seams (#36, #45, #46) become parity fixtures for Rust without simultaneously replacing rendering.

### B. Full Bevy 0.19 WASM/native runtime

Bevy owns ECS, rendering, input, asset lifecycle, fixed-step scheduling, and platform abstraction. Pair with Rapier if its measured physical behavior is suitable. Browser target should support WebGL2 as broad fallback and WebGPU as the high-end path.

[...]

PR #54, opened 2026-09-03
Define backend-neutral acoustic contact model

xtreemze:
Refs #53, #52
Related: #58, #59, #60, #61

## Ownership boundary

Pure acoustic-material/contact modeling only. No Babylon, Cannon, WAFXR, Tone, Web Audio, Bevy, Rapier, or Rust integration; no audible output changes.

## Change

Add `src/js/audio/acoustics.ts` with:

- semantic acoustic material IDs independent from visual `StandardMaterial` instances;
- provisional acoustic material profiles for towers, projectiles, enemies, ground, energy reserve, and fragments;
- a physics-facing `AcousticContact` contract carrying normal/tangential/angular motion, effective mass, body scale, position, material pair, damage energy, and deterministic seed;
- an `impactEnergy()` calculation aligned with #59 terrain impact energy;
- a backend-neutral `AcousticExcitation` descriptor derived from contact measurements plus explicit calibration values;
- finite-safe sanitization so malformed contact/calibration values cannot emit NaN/Infinity into DSP planning.

The derived descriptor includes impact energy, normalized energy, damage energy, normalized damage energy, modal fundamental, decay, brightness, noise/scrape mix, saturation, mode ratios, position, and seed. It deliberately contains no playback API.

## Material pair semantics

A collision's `materialA` / `materialB` ordering is physics-engine bookkeeping, not acoustic meaning. The blended material must therefore be commutative.

The model now:
- canonicalizes material order before combining mode lists;
- averages material coefficients symmetrically;
- does **not** invent a pair material id by reusing material A's id;
- avoids duplicating identical mode ratios for same-material contacts;
- returns a copied mode list so a backend cannot mutate the global material table through an excitation descriptor.

Local validation should prove `deriveAcousticExcitation(A,B) === deriveAcousticExcitation(B,A)` for every semantic material pair when all physical inputs are otherwise equal.

## Shared impact-energy convention

`normalSpeed` may now be signed or absolute. `impactEnergy()` uses its magnitude, matching #59 `terrainImpactEnergy()` so terrain deformation and acoustic excitation can consume the same signed contact-normal velocity without one subsystem silently treating an approaching contact as zero energy.

This should become part of the simulation/contact protocol used by #66 rather than being re-derived differently in each backend.

## Damage energy

`damageEnergy` was previously accepted by `AcousticContact` and then discarded. It is now preserved in `AcousticExcitation` as both raw `damageEnergy` and `normalizedDamageEnergy` using the same provisional impact-energy reference scale.

This PR deliberately does **not** use damage energy to retune brightness, saturation, pitch, or decay yet. Listening fixtures should determine whether damage should produce resonance instability/fracture coloration rather than baking an untested mapping into the first contract.

## Why

Defend's existing sound engine is procedural and positional, but its synthesis parameters are mostly selected by named events, level, HP, or object counts. #53 moves the design toward sound that is caused by the same physical quantities the player sees: impact energy, relative motion, body scale, materials, damage, and contact position.

This module is also intended as a portability seam: a future Web Audio implementation and a Rust/WASM DSP implementation can consume the same conceptual event/model, and Rapier can later source the same contract if #52/#66 advances.

## Important: provisional calibration

The material coefficients and mode ratios in this first slice are hypotheses for controlled listening tests, **not** established gameplay constants. They must not be treated as frozen balance or acoustic truth. The public contract and physical inputs are more important than these initial values.

## Validation required before promotion

Local Codex should:

1. confirm the historical TypeScript 3.2 toolchain accepts the module syntax;
2. property-test all public outputs across zero, negative, NaN, ±Infinity, and very large contact/calibration inputs; every scalar/vector component must remain finite;
3. verify positive and negative `normalSpeed` of equal magnitude produce identical impact energy;
4. verify swapping material A/B produces identical excitation values and identical mode ordering for every material pair;
5. verify same-material contacts do not duplicate their mode list;
6. verify `normalizedDamageEnergy` is bounded `[0,1]` and changes monotonically with raw positive damage energy;
7. confirm ordinary finite fixtures from the previous revision remain unchanged except for the intentionally added damage fields/material-pair ordering fixes;
8. build a small offline/browser synthesis harness that maps `AcousticExcitation` to a simple modal/noise voice without touching production sound;
9. compare projectile→enemy, enemy→ground, and enemy→energy material pairs;
10. identify calibration values that produce useful audible separation without excessive level variance;
11. verify the model can be mirrored straightforwardly in the dependency-light Rust core/runtime and that the TS/Rust excitation fixtures match.

## Gameplay/audio impact

None in this PR. The module is not wired to current sound call sites.

Keep as draft until the pure model and calibration assumptions have been reviewed with deterministic fixtures.

### DIFF
--- a/src/js/audio/acoustics.ts
+++ b/src/js/audio/acoustics.ts
@@ -1,0 +1,206 @@
+export type AcousticMaterialId =
+	| "tower"
+	| "projectile"
+	| "enemy"
+	| "ground"
+	| "energy"
+	| "fragment";
+
+export interface AcousticMaterial {
+	id: AcousticMaterialId;
+	density: number;
+	stiffness: number;
+	damping: number;
+	roughness: number;
+	brightness: number;
+	noise: number;
+	saturation: number;
+	modeRatios: number[];
+}
+
+export interface AcousticVector3 {
+	x: number;
+	y: number;
+	z: number;
+}
+
+export interface AcousticContact {
+	position: AcousticVector3;
+	normal: AcousticVector3;
+	normalSpeed: number;
+	tangentialSpeed: number;
+	angularSpeed: number;
+	effectiveMass: number;
+	bodyScale: number;
+	materialA: AcousticMaterialId;
+	materialB: AcousticMaterialId;
+	damageEnergy: number;
+	seed: number;
+}
+
+export interface AcousticCalibration {
+	referenceImpactEnergy: number;
+	referenceSpeed: number;
+	referenceBodyScale: number;
+	referenceFrequencyHz: number;
+	referenceDecaySeconds: number;
+}
+
+export interface AcousticExcitation {
+	position: AcousticVector3;
+	impactEnergy: number;
+	normalizedEnergy: number;
+	fundamentalHz: number;
+	decaySeconds: number;
+	brightness: number;
+	noiseMix: number;
+	scrapeMix: number;
+	saturation: number;
+	modeRatios: number[];
+	seed: number;
+}
+
+export const acousticMaterials: { [key: string]: AcousticMaterial } = {
+	tower: {
+		id: "tower",
+		density: 0.9,
+		stiffness: 0.85,
+		damping: 0.35,
+		roughness: 0.35,
+		brightness: 0.65,
+		noise: 0.2,
+		saturation: 0.25,
+		modeRatios: [1, 1.61, 2.37, 3.91]
+	},
+	projectile: {
+		id: "projectile",
+		density: 1,
+		stiffness: 1,
+		damping: 0.2,
+		roughness: 0.15,
+		brightness: 0.95,
+		noise: 0.15,
+		saturation: 0.45,
+		modeRatios: [1, 2.08, 3.23, 5.17]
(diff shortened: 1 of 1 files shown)


xtreemze (commented):
Online static review at head `78030fe883215becb7cb56dd92655e9d7829f5ff`: material-pair combination is now commutative, signed normal-speed impact energy aligns with terrain/Rust semantics, and damage energy is no longer silently discarded. No production imports are introduced. Keep draft until TS 3.2/property fixtures and listening/synthesis calibration evidence are posted.

xtreemze:
Safe-merge classification for current head `78030fe883215becb7cb56dd92655e9d7829f5ff`: this PR adds exactly one pure, currently unimported module (`src/js/audio/acoustics.ts`) and does not change any production call site or runtime behavior. Static inspection confirms it uses TypeScript 3.2-era syntax only. The numerical/property/listening checklist in the PR remains required **before this model is imported or wired into production audio**, but it no longer needs to block merging the dormant contract itself. Any future consumer PR must cite those fixture results.

Issue #56, opened 2026-09-03
Design a hex-centered arena with rhythmic spatial strategy


xtreemze:
Parent: #29
Related: #30, #32, #37, #47, #52, #53

## Direction

Evolve Defend from a square arena with square placement quantization into a six-fold, hex-centered spatial system while preserving the core interaction model: the camera continues to orbit the central energy reserve, enemies remain free physics bodies rather than pathfinding units, and the player still manipulates the battlefield directly.

The important design distinction is that the **hex lattice organizes construction, territory, perception, timing, and scenario topology; it does not become enemy navigation.** Physics remains authoritative.

## Current constraints observed in the repository

- `map.ts` creates one square `CreateGround` mesh sized from `mapGlobals.size`;
- tower placement snaps pointer hits independently on x/z to a 10-unit square lattice;
- the central reserve is a 40×40 box with sixteen manually enumerated blocked square cells around it;
- the ArcRotateCamera already orbits a center target and is therefore conceptually compatible with a radial/hex arena;
- enemy spawning already has an origin/sector concept that can be generalized into six radial approach sectors once #47 fixes occupancy ownership.

### 1. Hex energy core

Replace the square bank foundation with a regular hexagonal energy plinth centered at the world origin.

Treat the center as ring 0. Reserve ring 1 (or a configurable inner radius) as protected core space. This gives the defended objective a clear visual and tactical identity from every camera azimuth.

The energy reserve itself can remain physically simple during migration; the visual/placement topology can become hexagonal before collision geometry is made more complex.

### 2. Axial/cube coordinates as the logical placement system

Use axial `(q, r)` or cube `(q, r, s)` coordinates for all build cells and convert them to world x/z for rendering/physics.

This replaces:
- hand-written central blocked coordinates;
- square modulo snapping;
- implicit x/z equality as cell identity.

Cell identity should become an explicit value that can be shared by TypeScript and the Rust core in #52.

### 3. Concentric rings + six sectors

A regular hex grid naturally gives:
- concentric distance rings around the core;
- six principal approach sectors;
- six adjacency directions;
- meaningful radial vs tangential tower placement.

These should become strategy vocabulary. A tower can defend a sector, bridge adjacent sectors, form a ring, or intentionally leave a physical channel open for enemies to be redirected/ejected.

### 4. Scenario topology instead of one fixed board

Do not stop at a larger regular hex. Represent an arena as a set of enabled hex cells plus optional continuous boundary/collider geometry.

[...]

Issue #47, opened 2026-09-03
Certify and repair enemy spawn occupancy ownership


xtreemze:
Parent: #32
Related: #30, #37

## Observed legacy behavior

Enemy spawn occupancy currently has two independent correctness problems.

### 1. The occupancy branch in `enemyGenerator()` does not affect the selected location

`src/js/enemy/Enemy.ts` generates `newLocation`, checks whether it appears in both `enemyGlobals.occupiedSpaces` and the protected economy cells, but then calls `positionGenerator(waveOrigin)` again in **both** the `if` and the `else` branch.

As a result, the checked location is always discarded and the final location is never validated against either occupancy set.

### 2. Destroyed enemies release an unrelated reservation

Spawns record coordinates with:

```ts
enemyGlobals.occupiedSpaces.unshift([newLocation.x, newLocation.z]);
```

but `destroyEnemy()` releases occupancy with:

```ts
enemyGlobals.occupiedSpaces.pop();
```

No coordinate or enemy identity is stored on the enemy, so the reservation removed is simply the oldest array entry, not necessarily the position of the enemy being destroyed.

## Why this matters

Spawn placement affects wave geometry, collision density, central-bank pressure, tower value, and any deterministic physics benchmark. Before modernizing the scheduler or physics engine, we need to know the actual current spawn distribution and then replace this bookkeeping with explicit ownership.

`positionGenerator()` also reveals that `waveOrigin` controls a fixed x-origin (`-1`, `0`, or `1` map to distinct lanes/sides) while z is randomized, so a repaired allocator must preserve the intended wave-origin semantics rather than replacing it with generic random placement.

## Local Codex characterization

Against current `master`:

- instrument at least several representative waves containing default/undefined origin, origin `0`, and origin `1`;
- record generated x/z coordinates and collisions/duplicates among simultaneously live enemies;
- record whether any spawn overlaps the protected central economy cells;
- record `enemyGlobals.occupiedSpaces` length/content as enemies spawn and die out of order;
- determine whether duplicate coordinates are common enough to materially affect initial collision behavior;
- verify the exact lane/origin behavior intended by the fourth wave-array element.

## Implementation target

After baseline evidence is posted:

- generate a candidate and validate **that same candidate**;
- retry with an explicit bounded policy when a cell is occupied;
- associate each live enemy with its own spawn reservation (typed metadata or an explicit map keyed by entity identity);
- release exactly that reservation on destruction, idempotently;
- keep protected economy cells unavailable;
[...]

PR #57, opened 2026-09-03
Define pure hex topology primitives

xtreemze:
Refs #56, #52
Related: #30, #32, #47, #58, #59, #66

## Scope

Pure hex-grid topology only. No rendering, map mesh, tower placement, spawn behavior, camera, physics, audio, or balance changes.

## Change

Add `src/js/gameplay/hexGrid.ts` with a pointy-top axial-coordinate model:

- `HexCell` axial `(q, r)` and cube conversion;
- hex ↔ world x/z conversion;
- nearest-cell cube rounding for pointer/world hits;
- six canonical readonly neighbors;
- hex distance/ring calculation;
- configurable arena-radius membership;
- configurable protected core radius;
- deterministic six-sector classification around the central energy core;
- stable string cell keys;
- finite-safe handling of zero/invalid size, directions, coordinates and radii.

## Why

The current live game snaps x/z independently onto a 10-unit square lattice, hard-codes sixteen blocked central cells around a 40×40 square energy bank, and uses world x/z equality as cell identity. #56 proposes separating logical topology from continuous physics.

This module is the first low-risk seam for that migration. Hexes organize construction/territory/sector timing, while enemies/projectiles remain unconstrained continuous-world physics bodies.

It is intentionally dependency-free so the exact topology can be mirrored in the Rust `defend-core` / #67 runtime path.

## Orientation

This pass still uses a **pointy-top** orientation. Orientation remains provisional until the Storybook/browser fixture compares camera readability, tower footprint alignment, pointer ownership and arena masks. Changing orientation before production wiring remains cheap.

## Deterministic rounding rule

Cube rounding keeps the original draft's tie behavior but now documents it as protocol semantics. On exact equal rounding-error ties, correction prefers the later axis in q → r → s order (`s`, then `r`, then `q`).

This matters only on exact hex edges/vertices, but those are precisely the cases where TS/Rust implementations could otherwise disagree. The Rust mirror must use the same rule.

Logical operations canonicalize cells through this rounding rule before neighbor/distance/ring/key calculations, so accidental fractional coordinates do not create a second form of cell identity.

## Deterministic sectors without trigonometry

`hexSector()` no longer uses `atan2`. For pointy-top axial coordinates the six world-direction dot products reduce to exact linear scores:

- sector 0: `2q + r`
- sector 1: `q + 2r`
- sector 2: `-q + r`
- sector 3: `-2q - r`
- sector 4: `-q - 2r`
- sector 5: `q - r`

The highest score wins. Exact sector-boundary ties choose the lowest sector id because later equal scores do not replace the current winner.

Benefits:
- no floating-point angle normalization;
- no platform-specific trig boundary behavior;
- sector identity depends only on logical topology, not presentation scale;
- straightforward byte-for-byte TS/Rust parity;
- cheaper classification for wave/pressure/terrain/audio metadata.

The center remains sector `-1`.

## Numeric boundary

Zero/negative/non-finite hex sizes are normalized to a small positive magnitude before coordinate division. Invalid direction indices are normalized through finite integer modulo. Radius/protected-radius inputs are finite, nonnegative integers. Absurd/non-finite scalar state is saturated at an internal defensive ceiling so one malformed debug/input value cannot produce Infinity/NaN topology state.

These are robustness rules, not game-world size or balance limits.

## Readonly direction contract

`HEX_DIRECTIONS` is now both runtime-frozen and typed as `ReadonlyArray<Readonly<HexCell>>`. Consumers may enumerate/copy directions but should not mutate the canonical topology constants.

## Local validation before promotion

1. Confirm TypeScript 3.2 accepts the module, including `ReadonlyArray<Readonly<HexCell>>` and default parameters.
2. Property-check finite output for zero/negative/NaN/±Infinity size, coordinates, directions and radii.
3. Property-check round trips: cell → world → cell for a representative arena radius and several positive/negative sizes.
4. Verify exact edge/vertex fixtures follow the documented cube-round tie convention.
5. Verify every canonical cell has six unique reciprocal neighbors.
6. Verify direction wrapping for negative, >5 and non-integer direction values is deterministic.
7. Verify ring counts are `1` for radius 0 and `6r` cells on each ring `r > 0`.
8. Verify radius/protected-radius membership is invariant for equivalent canonical cells.
9. Verify the six axial sector directions map to sectors 0…5 respectively.
10. Verify exact sector-boundary cells choose the documented lowest-id tie winner and sector classification is independent of positive hex size.
11. Verify sector classification rotated by one canonical 60° hex direction advances coherently modulo 6.
12. Render a Storybook/browser overlay after #65 certification and compare pointy-top vs flat-top readability before wiring placement.
13. Mirror coordinate conversion, cube rounding, neighbors, distance/rings, sector scores/tie rules and key serialization in Rust; compare exhaustive fixtures for a bounded radius.
14. Use the same topology fixtures when #59 moves to six-neighbor terrain updates so neighbor order/topology does not drift between TS and Rust.

## Gameplay impact

None. This file is not imported by production code.

Keep as draft until topology fixtures and camera/interaction testing are posted.

### DIFF
--- a/src/js/gameplay/hexGrid.ts
+++ b/src/js/gameplay/hexGrid.ts
@@ -15,42 +15,84 @@ interface WorldPoint2D {
 }
 
 const SQRT_3 = Math.sqrt(3);
+const MIN_HEX_SIZE = 0.000001;
+const MAX_HEX_SCALAR = 1e100;
 
-const HEX_DIRECTIONS: HexCell[] = [
-	{ q: 1, r: 0 },
-	{ q: 1, r: -1 },
-	{ q: 0, r: -1 },
-	{ q: -1, r: 0 },
-	{ q: -1, r: 1 },
-	{ q: 0, r: 1 }
-];
+function finiteScalar(value: number, fallback = 0): number {
+	if (value !== value) {
+		return fallback;
+	}
+	if (value > MAX_HEX_SCALAR) {
+		return MAX_HEX_SCALAR;
+	}
+	if (value < -MAX_HEX_SCALAR) {
+		return -MAX_HEX_SCALAR;
+	}
+	return value;
+}
+
+function positive(value: number): number {
+	return Math.max(0, finiteScalar(value));
+}
+
+function hexSize(size: number): number {
+	return Math.max(MIN_HEX_SIZE, Math.abs(finiteScalar(size, 1)));
+}
+
+function normalizedDirection(direction: number): number {
+	const integerDirection = Math.floor(finiteScalar(direction));
+	return ((integerDirection % 6) + 6) % 6;
+}
+
+const HEX_DIRECTIONS: HexCell[] = Object.freeze([
+	Object.freeze({ q: 1, r: 0 }),
+	Object.freeze({ q: 1, r: -1 }),
+	Object.freeze({ q: 0, r: -1 }),
+	Object.freeze({ q: -1, r: 0 }),
+	Object.freeze({ q: -1, r: 1 }),
+	Object.freeze({ q: 0, r: 1 })
+]) as HexCell[];
 
 function hexToCube(cell: HexCell): HexCube {
+	const q = finiteScalar(cell.q);
+	const r = finiteScalar(cell.r);
 	return {
-		q: cell.q,
-		r: cell.r,
-		s: -cell.q - cell.r
+		q,
+		r,
+		s: finiteScalar(-q - r)
 	};
 }
 
 function cubeToHex(cube: HexCube): HexCell {
-	return { q: cube.q, r: cube.r };
+	return { q: finiteScalar(cube.q), r: finiteScalar(cube.r) };
 }
 
 function hexToWorld(cell: HexCell, size: number): WorldPoint2D {
+	const safeSize = hexSize(size);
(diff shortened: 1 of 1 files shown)


xtreemze (commented):
Online static review at head `c1a21216a2ef8c20c1c803a27c4efc1650768d23`: topology is now dependency-free, finite-safe, explicit about cube-round boundary ties, and sector classification uses deterministic axial linear scores rather than trig. Exact-boundary tie choices are intentional protocol semantics, not evidence that pointy-top orientation is final. Keep draft until exhaustive TS/Rust topology fixtures plus camera/pointer Storybook evaluation are posted.

xtreemze:
Safe-merge classification for current head `c1a21216a2ef8c20c1c803a27c4efc1650768d23`: this PR adds exactly one pure, currently unimported topology module (`src/js/gameplay/hexGrid.ts`) and changes no placement, map, spawn, camera, physics, audio, or balance behavior. Static inspection confirms TS 3.2-era syntax only. The round-trip/property/Storybook/Rust-parity checklist remains mandatory **before production placement/topology code imports this module**; those fixtures no longer need to block merging the dormant primitive itself.

Issue #58, opened 2026-09-03
Add bounded self-leveling terrain deformation from physical impacts


xtreemze:
Parent: #29
Related: #32, #47, #52, #53, #56, #57

## Goal

Let heavy enemy bodies and high-energy impacts leave shallow, temporary deformation in the arena so the world visibly remembers physical stress. Nearby player structures, walls, the energy core, and the camera target should settle with the supporting ground rather than remain visually detached from it.

This must **not** become a crater/sinkhole mechanic that traps enemies or permanently destroys navigability. The terrain should continuously relax toward a stable level, with stronger/faster stabilization around maintained construction.

## Design principle

Treat terrain deformation as a **bounded support-height field** driven by physics, not as unconstrained destruction of the collision mesh.

The support field can be shared by:
- visual ground displacement;
- structure/foundation support height;
- energy-core/plinth support height;
- camera-target vertical offset;
- future acoustic surface response in #53;
- optionally a lower-frequency/coarser collision heightfield once a physics backend can support it safely.

Enemies/projectiles remain continuous-world physics bodies. The hex topology from #56/#57 may index deformation samples, but terrain must not constrain movement to hexes.

## Impact model

An impact should deposit deformation according to physically meaningful values where available:

- effective mass;
- normal impact speed / normal impulse;
- contact position;
- body radius / projected contact area;
- ground/material compliance;
- existing local deformation;
- deterministic seed only for very small visual variation, never for core displacement magnitude.

Use saturating mappings so heavier enemies mainly create **wider** depressions rather than arbitrarily deeper ones.

Suggested behavior:

- impact energy controls total deformation energy;
- body scale/contact radius controls footprint radius;
- depth uses a capped nonlinear response (`tanh`, logarithmic, or equivalent);
- slope is explicitly limited;
- repeated impacts accumulate with diminishing returns near the local depth cap;
- optional very small displaced-material rim/heave may appear around the depression, but should recover faster than the central dent.

## Anti-trapping invariants

These are hard gameplay constraints:

1. maximum depression depth is small relative to the smallest enemy diameter;
2. maximum local slope is bounded below a value likely to trap rolling enemies;
3. repeated impacts cannot deepen terrain without bound;
4. deformation diffuses spatially as well as relaxing vertically;
5. collision terrain, if enabled, is smoother/coarser than visual deformation;
6. the physical collision floor may remain flatter than the rendered/support field in the first implementation;
7. terrain recovery continues even under sustained load, though active pressure may slow it;
[...]

PR #55, opened 2026-09-03
Spike a Rust/WASM gameplay core

xtreemze:
Refs #52
Related: #30, #36, #45, #46, #53

## Scope

Experimental Rust/WASM parity spike only. This does **not** replace Babylon, Cannon, TypeScript gameplay, rendering, input, or audio, and it is not intended to merge until the local executor proves the build/interop path is worthwhile.

## Change

- add a minimal Cargo workspace with `crates/defend-core`;
- keep the crate dependency-light: only optional `wasm-bindgen` support;
- expose current characterized projectile, turret, enemy, economy, and acoustic-impact calculations as pure Rust functions;
- add native unit tests for the exact current numeric contracts already documented in #36/#45/#46;
- allow the same functions to be exported to browser JavaScript when built with the `wasm` feature.

No Bevy or Rapier dependency is introduced in this first spike. The goal is to establish a small, testable Rust boundary before evaluating a larger engine migration.

## Mirrored contracts

Projectile/turret:
- L2 HP 1760, impulse 26560, mass 120, cadence 208 ms;
- L3 HP 5940, impulse 89640, mass 270, cadence 702 ms.

Enemy:
- diameters 6 / 9 / 14;
- HP 15440 / 60880 / 136320;
- mass 5400 / 21600 / 48600;
- movement impulse 7400 / 29600 / 66600;
- current legacy restitution -0.02 / -0.32 / -0.82.

Economy:
- projectile recovery uses current hit-points × ratio with max-balance cap;
- bank damage remains remaining enemy HP / 2.

Audio bridge:
- includes only the physically meaningful scalar `impact_energy(effective_mass, normal_speed)` as a first shared primitive for #53.

## Local Codex certification

Please validate this exact head before considering promotion:

1. `cargo test -p defend-core` on the current stable Rust toolchain;
2. `cargo check -p defend-core --features wasm --target wasm32-unknown-unknown`;
3. build browser bindings with a current `wasm-bindgen`/`wasm-pack` workflow and report generated WASM + JS glue sizes (raw and compressed where practical);
4. call several exported functions from a tiny browser harness and verify numeric parity with the current TypeScript formulas;
5. record compile time, incremental compile time, debugging/source-map ergonomics, and any browser/PWA integration friction;
6. evaluate whether the 2024 edition or optional dependency choices create avoidable compatibility constraints;
7. do not add Bevy/Rapier until this lower-level parity boundary is understood.

## Decision gate

[...]

PR #59, opened 2026-09-03
Define bounded terrain deformation primitives

xtreemze:
Refs #58
Related: #52, #53, #56, #57, #66

## Scope

Pure terrain-deformation math only. No production ground mesh mutation, collision changes, structure movement, camera movement, audio output, or gameplay balance changes.

## Change

Add `src/js/gameplay/terrainDeformation.ts` with renderer/physics-independent primitives for:

- physically meaningful normal-impact energy;
- bounded impact depth and footprint calculation;
- smooth radial deformation deposition;
- repeated-impact accumulation with diminishing returns at a hard depth cap;
- finite-safe handling of malformed/extreme numeric state;
- structure stabilization weighting;
- frame-rate-independent vertical recovery toward baseline;
- lateral diffusion toward neighboring terrain samples;
- pairwise slope limiting plus order-independent multi-neighbor slope bounds;
- damped/capped support-height settling for towers, walls, the core, and camera target;
- calibration-aware support settling using `maxSupportSpeed`.

## Important behavior

The model intentionally makes mass/energy broaden the affected area more readily than deepen it. Maximum depth is always explicit and repeated impacts cannot accumulate without bound.

The functions do not assume a square or hex grid. A future hex field can pass the mean of its six neighbors to `diffuseDepression()` and the same six-neighbor snapshot to `limitDepthByNeighborSlopes()`.

`terrainImpactEnergy()` uses the same scalar physical concept as #53's acoustic work so terrain and sound may react coherently to one collision description without either subsystem becoming authoritative over the other.

## Deterministic slope rule

Sequentially mutating one sample against neighbor A, then feeding that result into neighbor B, makes slope limiting depend on neighbor iteration order. That would undermine deterministic replay and make future six-neighbor hex updates fragile.

The module now exposes:
- `terrainDepthBoundsFromNeighbors()` to compute the allowed depth interval from one immutable neighbor snapshot;
- `limitDepthByNeighborSlopes()` to apply those bounds once, independent of neighbor ordering;
- `limitDepthBySlope()` retained as a pairwise compatibility wrapper.

If the neighboring snapshot is itself inconsistent (no single depth can satisfy every pairwise slope cap), the multi-neighbor helper returns the midpoint of the conflicting bounds. This deterministic minimax fallback reduces the worst symmetric violation; subsequent snapshot-based diffusion/slope passes are expected to converge. It does not pretend an impossible local constraint set was satisfied.

## Numeric boundary

Pure terrain helpers now sanitize malformed values and saturate absurd scalar magnitudes at an internal defensive ceiling. Ordinary game values are vastly below this ceiling. The goal is only to guarantee finite state propagation when a physics/contact/debug input is briefly invalid; this is not a gameplay/world-size limit.

## Support settlement

`settleSupportHeight()` remains the generic primitive with an explicit speed cap. `settleCalibratedSupportHeight()` now routes `TerrainDeformationCalibration.maxSupportSpeed` through that primitive so the calibration field is not disconnected from the model.

## Not included

- no Babylon mesh edits;
- no Cannon heightfield/trimesh work;
- no Rapier dependency;
- no calibration constants wired into production;
- no automatic tower/core/camera settling yet;
- no visual crater effects;
- no persistence/serialization wiring.

## Local validation before promotion

1. Confirm TypeScript 3.2 accepts the file.
2. Property-test every public function across zero/extreme/NaN/±Infinity mass, speed, compliance, stabilization, timing, slope and support inputs; outputs must remain finite.
3. Verify depth is monotonic with energy but never exceeds `maxDepth`.
4. Verify increasing body radius/energy broadens footprint while depth remains saturated.
5. Repeated deposits at one point must asymptotically approach, never exceed, `maxDepth`.
6. Recovery should respect configured half-life independent of frame step.
7. Stabilized samples should recover faster but remain continuous.
8. Diffusion must reduce sharp local differences without producing negative depth.
9. `settleSupportHeight()` and `settleCalibratedSupportHeight()` must obey maximum vertical speed and not snap on large target changes.
10. For a consistent neighbor snapshot, verify `limitDepthByNeighborSlopes()` is invariant under all neighbor permutations and satisfies every pairwise slope cap.
11. For an inconsistent neighbor snapshot, verify the result is permutation-invariant, bounded, and equals the midpoint of the conflicting aggregate bounds.
12. Run repeated six-neighbor snapshot-based slope + diffusion passes and verify convergence/no oscillatory order artifact on sharp synthetic pits/ridges.
13. Mirror these pure functions in the Rust `defend-core`/runtime path once #55/#67 are ready, then compare deterministic fixtures.
14. Build the Storybook terrain playground after #65 is certified; visualize raw deposits, aggregate slope bounds, diffusion/recovery, stabilization and support settlement separately.

## Gameplay impact

None in this PR. This module is not imported by production code.

Keep as draft until local numeric/property fixtures are posted and initial deformation calibration is informed by browser physics experiments.

### DIFF
--- a/src/js/gameplay/terrainDeformation.ts
+++ b/src/js/gameplay/terrainDeformation.ts
@@ -24,14 +24,36 @@ export interface TerrainImpactProfile {
 	footprintRadius: number;
 }
 
+export interface TerrainDepthBounds {
+	minimum: number;
+	maximum: number;
+	/** False means the supplied neighbor snapshot already violates the slope cap. */
+	consistent: boolean;
+}
+
 const LN2 = Math.log(2);
+const MAX_TERRAIN_SCALAR = 1e100;
+
+function finiteScalar(value: number, fallback = 0): number {
+	if (value !== value) {
+		return fallback;
+	}
+	if (value > MAX_TERRAIN_SCALAR) {
+		return MAX_TERRAIN_SCALAR;
+	}
+	if (value < -MAX_TERRAIN_SCALAR) {
+		return -MAX_TERRAIN_SCALAR;
+	}
+	return value;
+}
 
 function clamp(value: number, minimum: number, maximum: number): number {
-	return Math.max(minimum, Math.min(maximum, value));
+	const safeValue = finiteScalar(value, minimum);
+	return Math.max(minimum, Math.min(maximum, safeValue));
 }
 
 function positive(value: number): number {
-	return Math.max(0, value);
+	return Math.max(0, finiteScalar(value, 0));
 }
 
 /**
@@ -43,7 +65,9 @@ export function terrainImpactEnergy(
 	effectiveMass: number,
 	normalSpeed: number
 ): number {
-	return 0.5 * positive(effectiveMass) * Math.pow(Math.abs(normalSpeed), 2);
+	const mass = positive(effectiveMass);
+	const speed = Math.abs(finiteScalar(normalSpeed));
+	return positive(finiteScalar(0.5 * mass * Math.pow(speed, 2)));
 }
 
 /**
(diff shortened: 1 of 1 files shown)


xtreemze (commented):
Online static review at head `064cba399f807df6892953cb716bbe3165a2399e`: the terrain seam remains topology/renderer/physics independent, finite-safe, depth-capped, and now supports immutable-snapshot multi-neighbor slope bounds instead of order-dependent sequential mutation. The deterministic midpoint fallback for inconsistent local constraints is explicitly provisional and must be stress-tested for convergence. Keep draft until property/numeric and browser anti-trapping fixtures are posted.

xtreemze:
Safe-merge classification for current head `064cba399f807df6892953cb716bbe3165a2399e`: this PR adds exactly one pure, currently unimported terrain math module (`src/js/gameplay/terrainDeformation.ts`) and does not mutate Babylon ground, Cannon collision, structures, camera, audio, or gameplay. Static inspection confirms TS 3.2-era syntax only. The finite/property/convergence/Storybook/Rust-parity checklist remains mandatory **before any terrain state, mesh, support-height, camera, or collision integration consumes this module**; it no longer blocks merging the dormant primitive.

PR #61, opened 2026-09-03
Define object-based spatial audio primitives

xtreemze:
Refs #60
Related: #52, #53, #56, #58

## Scope

Pure spatial-audio scene math only. No production sound routing, WAFXR/Tone changes, AudioWorklet, Worker, visualization, listener wiring, or gameplay behavior changes.

## Change

Add `src/js/audio/spatialAudio.ts` with backend-neutral primitives for:

- world-space emitter objects with stable id, acoustic profile, position, velocity, orientation/directivity, size, excitation, threat and continuity state;
- listener position, velocity and orientation;
- source/listener distance;
- bounded future closest-approach prediction for fly-by prioritization;
- explicit Doppler ratio from relative radial source/listener velocity;
- distance-gain hints with a preserved near field;
- distance-dependent high-frequency retention / air-loss hints;
- deterministic perceptual/gameplay priority combining proximity, predicted close pass, excitation energy, threat and sustained-voice continuity;
- one aggregate `spatialRenderHints()` output for renderer backends;
- defensive numeric sanitization so malformed/non-finite emitter or calibration state cannot poison a complete planning pass.

## Why

The current WAFXR path already positions many one-shot sounds using `soundX/Y/Z`, and `arcCamera.ts` updates the listener each frame. However source positions are sampled only when sounds start, source/listener velocity is not represented, current Web Audio panners do not provide automatic velocity/Doppler behavior, and the existing voice budget is timer/counter based.

This PR establishes a scene-object contract before changing the renderer.

## Numeric boundary

Spatial/audio planning must remain finite even when an upstream transform, physics event, worker message, debug control, or calibration briefly contains invalid state.

The pure math now:
- maps `NaN` to neutral scalar values;
- saturates infinite/absurd scalar magnitudes at an internal `1e150` defensive ceiling;
- performs vector subtraction, scaling, dot products and closest-approach projection through the same finite boundary;
- clamps distance gain, high-frequency retention, energy score and final priority to their semantic `[0,1]` ranges;
- keeps all ordinary Defend world values many orders of magnitude below the defensive ceiling, so normal physical behavior is unchanged.

This is robustness, not gameplay calibration. The ceiling is not a world-size design limit and should not appear in UI or simulation rules.

## Doppler convention

The line-of-sight vector runs from listener to source.

- positive listener radial velocity toward the source raises pitch;
- positive source radial velocity away from the listener lowers pitch;
- tangential velocity has little instantaneous contribution;
[...]

Issue #60, opened 2026-09-03
Build an object-based spatial audio runtime with motion, Doppler, prioritization, and threaded rendering


xtreemze:
Parent: #29
Related: #52, #53, #56, #58

## Goal

Treat game audio as a set of world-space sound objects rather than one-shot effects. Each important emitter should carry its own 3D position, velocity/motion, orientation/directivity, physical/acoustic identity, lifetime and importance. The listener is the active camera/viewpoint and has its own position, orientation and velocity.

The same object model must work with the current web client and remain portable to a future Rust/WASM or Bevy/native backend.

## Current repository baseline

- `sound.ts` already passes static `soundX/Y/Z` values into WAFXR for many events.
- WAFXR creates a Tone `Panner3D` and sets the panner position only at sound start.
- `arcCamera.ts` updates listener position/orientation every render.
- there is no source velocity, listener velocity, Doppler, continuously moving emitter lifecycle, distance-aware voice prioritization, occlusion model, or worker/worklet separation.
- current voice limits use timer counters rather than perceptual/physical priority.

## Object contract

Define a backend-neutral `SpatialAudioObject`/emitter state containing at minimum:

- stable emitter id;
- source/event kind;
- world position;
- linear velocity;
- optional angular velocity;
- orientation/directivity;
- physical radius/extent;
- acoustic material/source profile;
- base gain / excitation energy;
- start time + lifetime / sustained state;
- threat/gameplay importance;
- deterministic variation seed;
- optional occlusion/transmission state later.

Define a `SpatialListenerState` containing:
- position;
- velocity;
- forward/up vectors;
- optional focus/core position for gameplay-priority calculations.

Transient collisions can create short-lived objects. Projectiles, rolling/scraping enemies and other sustained movers can own continuously updated emitter objects.

## Doppler and motion

Do not depend on automatic browser Doppler. Current Web Audio `PannerNode` exposes source position/orientation and distance behavior but no source/listener velocity API.

Compute Doppler explicitly from relative radial velocity between source and listener, with:
- configurable speed of sound in world units/second;
- physically bounded ratio;
- smoothing to avoid zippering;
- application to procedural oscillator phase/frequency or playback rate as appropriate;
- source and listener velocity both contributing;
- no Doppler for purely UI/non-world sounds.

A projectile passing close to a static camera should produce a convincing moving whoosh and pitch trajectory even before impact. Fast tangential movement should produce little Doppler until its radial component changes.

## Distance, air and salience

[...]

PR #62, opened 2026-09-03
Plan spatial audio quality tiers and virtualization

xtreemze:
Refs #60
Depends on #61

## Scope

Pure deterministic voice-budget planning only. This stacked PR adds no sound rendering, WAFXR/Tone changes, Worker, AudioWorklet, HRTF selection, visualization, or gameplay changes.

## Change

Add `src/js/audio/spatialVoiceBudget.ts` on top of #61:

- semantic renderer tiers: `full`, `standard`, `cheap`, `virtual`;
- deterministic ranking from #61 `spatialRenderHints()`;
- configurable counts for each rendered tier;
- explicit raw `priority` plus history-adjusted `selectionPriority` for diagnostics;
- stable emitter-id tie breaking for deterministic results;
- prototype-safe history state for successive planning passes;
- virtual voices remain semantically alive while consuming no rendering tier.

## Retention model

`activeVoiceRetention` is now a **rendered/virtual cutoff margin**, not an additive score bonus.

The planner first computes the raw priority ordering with no history. If there is a rendered/virtual boundary, a voice that was rendered on the previous pass may be retained only when its raw priority is below but within the configured margin of that cutoff. Its selection score is lifted only to the cutoff.

Consequences:

- a new source whose raw priority is above the cutoff always stays above retained boundary voices;
- history cannot globally boost stale voices above materially more important projectile fly-bys/core threats;
- hysteresis is limited to the audible rendered/virtual transition where churn matters most;
- `full`/`standard`/`cheap` quality transitions do not yet use history and should be made perceptually smooth by renderer-side crossfades/interpolation.

The retention margin remains capped at 0.25.

## Research rationale

Current Web Audio exposes both `equalpower` and `HRTF` spatialization. HRTF is the richer stereo localization mode, but the appropriate number of simultaneous HRTF/material-DSP voices should be measured per browser/device rather than hard-coded globally. The allocator therefore chooses abstract quality tiers only; a renderer later maps those tiers to HRTF/equal-power, modal count, update frequency, filters, or virtualization.

AudioWorklet is the preferred low-latency custom-DSP thread, while Worker is suitable for control/planning work. Keeping the planner pure means it may run on the main thread initially, in a Dedicated Worker, in Rust/WASM, or as part of a future native/Bevy path without changing semantics.

Primary references recorded in #60:
- Web Audio 1.1: https://webaudio.github.io/web-audio-api/
- PannerNode panning model: https://developer.mozilla.org/en-US/docs/Web/API/PannerNode/panningModel
- AudioWorklet: https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet
- Chrome AudioWorklet/WASM/Worker patterns: https://developer.chrome.com/blog/audio-worklet-design-pattern

## Local certification before promotion

1. Confirm TypeScript 3.2 accepts #61 + this file.
2. Generate 100+ deterministic emitters and verify plan output is repeatable across runs.
3. Verify tier counts exactly respect the configured budget.
4. Verify a high-priority newcomer above the raw cutoff cannot be displaced by a retained lower-priority voice.
5. Verify a previously rendered voice just below the cutoff can remain rendered when within `activeVoiceRetention`.
6. Verify a previously rendered voice outside the retention margin receives no history advantage.
7. Verify `priority` remains raw semantic priority while `selectionPriority` reflects boundary hysteresis only.
8. Verify `__proto__` or other unusual emitter ids cannot corrupt history state.
9. Measure planning cost for 100 / 500 / 1000 emitters on representative browsers; move planning to a Worker only if profiling justifies the messaging overhead.
10. Build the #63 fly-by fixture with several budgets and compare HRTF/equal-power CPU cost and localization quality before assigning concrete renderer behavior to tiers.

## Gameplay/audio impact

None. This PR is not connected to production audio.

Keep stacked/draft until #61 is certified and the deterministic stress fixtures are posted.

### DIFF
--- a/src/js/audio/spatialAudio.ts
+++ b/src/js/audio/spatialAudio.ts
@@ -67,39 +67,79 @@ export interface SpatialRenderHints {
 	priority: number;
 }
 
+/**
+ * Defensive numeric ceiling for this pure boundary. Real Defend world-space
+ * values are many orders of magnitude smaller; the ceiling only prevents
+ * malformed Infinity/NaN or absurd external state from poisoning an audio
+ * planning pass with non-finite output.
+ */
+const MAX_SPATIAL_SCALAR = 1e150;
+
+function finiteScalar(value: number, fallback = 0): number {
+	if (value !== value) {
+		return fallback;
+	}
+	if (value > MAX_SPATIAL_SCALAR) {
+		return MAX_SPATIAL_SCALAR;
+	}
+	if (value < -MAX_SPATIAL_SCALAR) {
+		return -MAX_SPATIAL_SCALAR;
+	}
+	return value;
+}
+
 function clamp(value: number, minimum: number, maximum: number): number {
-	return Math.max(minimum, Math.min(maximum, value));
+	const safeValue = finiteScalar(value, minimum);
+	return Math.max(minimum, Math.min(maximum, safeValue));
 }
 
 function positive(value: number): number {
-	return Math.max(0, value);
+	return Math.max(0, finiteScalar(value, 0));
 }
 
 function subtract(a: SpatialVector3, b: SpatialVector3): SpatialVector3 {
 	return {
-		x: a.x - b.x,
-		y: a.y - b.y,
-		z: a.z - b.z
+		x: finiteScalar(finiteScalar(a.x) - finiteScalar(b.x)),
+		y: finiteScalar(finiteScalar(a.y) - finiteScalar(b.y)),
+		z: finiteScalar(finiteScalar(a.z) - finiteScalar(b.z))
 	};
 }
 
 function dot(a: SpatialVector3, b: SpatialVector3): number {
-	return a.x * b.x + a.y * b.y + a.z * b.z;
+	return finiteScalar(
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Online static/stack review at head `d8ff447bfdad7501f1550dae68106396a40b4f07`: dependency graph is now correct—this branch contains current #61 head `4b9c41e` as a merge parent and the PR diff relative to #61 is only `spatialVoiceBudget.ts`. Boundary hysteresis no longer globally boosts stale voices. Keep draft until deterministic stress/property and browser cost/listening fixtures are posted.

xtreemze:
Safe-merge classification for normalized current head `29900f2a6adbb731993b03e19be0e77f10e7a9c2`: after merging #61, this branch was brought forward with a non-force two-parent merge whose tree is current `master` plus exactly one new file, `src/js/audio/spatialVoiceBudget.ts`. GitHub now reports that single-file diff. Static inspection confirms TS 3.2-era syntax only and no production imports. Deterministic 100/500/1000-emitter, hysteresis, HRTF/equal-power and Worker-placement fixtures remain mandatory before a runtime renderer consumes this allocator.

Issue #77, opened 2026-09-03
Prototype conserved teal energy spill and delayed silo collection


xtreemze:
Parent: #76
Related: #29, #30, #36, #75

## Purpose

Implement the first narrow proof of concept for physical teal-energy transfer without yet attempting full fluid simulation, mothership siphoning, world ecology, or final production balance.

## Ownership boundary

This issue owns only the **defender-side hit → visible spill → delayed silo collection** prototype.

It should stack on PR #36's economy seam rather than duplicating its arithmetic. Do not change:

- projectile damage formulas;
- enemy HP formulas;
- bank-damage semantics;
- tower balance;
- wave logic;
- mothership/raider extraction;
- final fluid rendering architecture;
- final transfer rates.

## PoC contract

Opt-in only, via a simple runtime/query flag so default gameplay remains observationally identical.

When enabled:

1. a projectile collision computes the same recovered-energy amount as the current economy rule;
2. that value is divided into a small bounded number of authoritative energy droplets/packets;
3. packets spawn around the enemy collision position;
4. packets settle toward the arena surface;
5. packets then stream across the surface toward the silo;
6. silo balance increases only as individual packets reach the silo;
7. total credited value must equal the baseline recovered value subject only to the existing max-balance cap;
8. visual particle count must remain small and bounded.

## Why this slice

This validates the most important new design claim before building richer liquid rendering:

**energy ownership changes only after visible spatial transfer.**

If this does not feel good at low complexity, adding metaballs, ribbons, GPU particles, or flow fields will not fix the game-design problem.

## Suggested implementation

- stack branch on #36 head `557dba7285d4d5152f3837db9ad11db2f0f295ba`;
- add an isolated `src/js/experiments/energyFlowPrototype.ts` module;
- reuse `projectileEnergyRecovery()` from #36;
- keep default instant credit path when the experiment is disabled;
- enable with `?energyFlow=1`;
- one scene-level render/update loop, not one callback per droplet;
- deterministic fixed spill offsets rather than unbounded randomness;
- Babylon meshes/materials are presentation only; economic value lives on the small authoritative packet records.

## Local Codex certification

Please validate the eventual PR head with:

1. historical build/TypeScript compatibility;
2. default mode has unchanged instant recovery;
3. `?energyFlow=1` delays credit until visible packets reach the silo;
4. T2 hit recovers exactly the same total as baseline (subject to cap);
5. T3 hit recovers exactly the same total as baseline (subject to cap);
[...]

Issue #76, opened 2026-09-03
Make teal energy a physical fluid and use energy signatures for the strategic world


xtreemze:
Parent: #73
Related: #29, #72, #74, #75

## Purpose

Replace abstract/instant energy transfer with a physically legible energy-flow system and extend that same visual language into raid extraction, raider evacuation, mothership target selection, and the wider silo/mothership ecology.

### Defensive recovery

- Projectile damage should no longer immediately teleport recovered teal energy into the silo.
- Damage liberates a conserved amount of teal energy at the collision.
- Energy falls/spills onto the arena as stylized liquid-like droplets.
- Droplets pool, ooze, merge, and form rivulets/streams.
- Free energy is gradually attracted toward the defending silo.
- The silo reserve increases only when the represented energy actually reaches the intake.
- The battlefield should visibly carry the economic result of combat home.

### Raid breach and extraction

When a raider reaches the silo:

- the direction of teal energy flow reverses;
- an extraction thread grows into a liquid-like upward stream from silo to mothership;
- stored energy visibly drains upward rather than disappearing numerically;
- remaining raiders disengage and use the same stream to ascend back to the mothership;
- raiders are physically entrained/lifted rather than teleported;
- the mothership visibly brightens as energy arrives.

Recommended experiment: allow defenders to continue firing during at least the early extraction/evacuation phase so reaching the silo is not necessarily an instantaneous binary end-state.

### Post-raid choice

After extraction/evacuation, the mothership should be able to:

1. return to / establish a defensive silo state; or
2. remain mobile and raid another energy source.

This supports a repeating lifecycle rather than a one-time story-mode inversion:

`SILO → DEFEND → DETER → STARVE → MOTHERSHIP → RAID → SETTLE/DEFEND → ...`

## Strategic world as teal signatures

Other silos should appear from the mothership as distant teal points/signatures rather than conventional mission cards or map nodes.

### Bright signatures

Bright teal sources generally imply:

- high stored energy and/or recent energy throughput;
- frequent raid activity;
- defenders that have had continuing opportunities to recover energy through combat;
- maintained, dense or battle-hardened defenses;
- high visibility to other raiders;
- high reward and high difficulty.

### Dim signatures

Dim sources generally imply:

- low stored energy;
- low recent raid activity;
- little combat-derived replenishment;
- defenses that have had more time to decay into disrepair;
- lower extraction reward;
- generally easier raids.

The intended readable trade-off is:

[...]

PR #78, opened 2026-09-03
experiment: prototype visible energy spill and silo collection

xtreemze:
Refs #77, #76, #86, #30
Stacked on #36 (`agent/gameplay-economy-seam`).

## Purpose

Prototype visible, conserved defender-side energy recovery with **real surface obstruction**:

`hit → bounded teal packets → fall/settle → obstacle-aware surface flow/pooling → silo credit`

Default gameplay remains unchanged unless `?energyFlow=1` is present.

## Current experiment behavior

`src/js/experiments/energyFlowPrototype.ts` now:

- keeps a bounded authoritative packet set (max 64 meshes);
- preserves #36 recovery arithmetic and the existing 30000 reserve cap;
- counts in-transit value against silo capacity;
- emits deterministic low-poly teal packets at the hit location;
- moves packets through a falling phase before surface collection;
- treats tagged `towerBase` footprints as physical flow obstacles;
- uses bounded local steering to find nearby openings rather than intelligent global maze solving;
- flattens blocked packets into a visible `pooling` state;
- keeps pooled value conserved and uncredited while no physical route exists;
- automatically resumes flow when tower/wall topology opens;
- exposes `activePackets`, `inTransitEnergy`, `poolingPackets`, `blockedEnergy`, and `routeChanges` via `window.__defendEnergyFlowPrototype.stats()`.

This implements the #86 principle that the silo attraction field biases energy but does **not** grant obstacle immunity. A sealed fortress can therefore strand its own recovered energy outside the silo.

## Conservation

Capacity is reserved using:

`maxBalance - currentBalance - inTransitEnergy`

Visual packet count never determines economic value. When the packet budget is exhausted, new recovered value merges into an existing aggregate instead of being lost.

## Scope boundary

This PR still does not implement mothership siphoning, raider evacuation, strategic signatures, final fluid rendering, production terrain flow fields, raider AI, bank-damage changes, or balance changes.

## Online/static review

Compared with parent #36 head `557dba7285d4d5152f3837db9ad11db2f0f295ba`:

- 4 commits ahead / 0 behind;
- 2 changed files;
- 532 additions / 7 deletions;
- no tower/enemy/wave/dependency changes.

## Local Codex certification required

1. historical TypeScript/build compatibility;
2. default URL still credits recovery immediately and matches #36;
3. `?energyFlow=1` visibly delays credit until arrival;
4. T2/T3 recovery remains exactly 246.4 / 831.6 when capacity permits;
5. near-full simultaneous hits never exceed 30000;
6. no double-credit/leak under repeated hits;
7. open-field spill reaches the silo;
8. a single tower/wall footprint diverts flow around its edge;
9. a U-shaped or sealed layout causes visible pooling rather than obstacle traversal;
10. opening/removing a blocker releases pooled value without creating or deleting energy;
11. inspect `poolingPackets`, `blockedEnergy`, route changes, frame time and mesh count under several simultaneous hits;
12. capture representative open-route, blocked, and released-flow evidence.

Keep draft until runtime evidence is posted.

### DIFF
--- a/src/js/experiments/energyFlowPrototype.ts
+++ b/src/js/experiments/energyFlowPrototype.ts
@@ -1,0 +1,282 @@
+import {
+	Color3,
+	Mesh,
+	MeshBuilder,
+	Scene,
+	StandardMaterial,
+	Vector3
+} from "babylonjs";
+import { economyGlobals } from "../main/globalVariables";
+
+type EnergyFlowPhase = "falling" | "streaming";
+
+interface EnergyFlowPacket {
+	value: number;
+	mesh: Mesh;
+	phase: EnergyFlowPhase;
+	verticalVelocity: number;
+	age: number;
+	wobble: number;
+}
+
+const MAX_ACTIVE_PACKETS = 64;
+const MIN_PACKETS_PER_HIT = 3;
+const MAX_PACKETS_PER_HIT = 6;
+const GROUND_Y = 0.42;
+const INTAKE_RADIUS = 3.2;
+const FALL_GRAVITY = 17;
+const BASE_STREAM_SPEED = 8;
+
+const SPILL_OFFSETS = [
+	new Vector3(-0.8, 0.4, -0.4),
+	new Vector3(0.7, 0.8, -0.2),
+	new Vector3(-0.3, 1.0, 0.7),
+	new Vector3(0.9, 0.5, 0.6),
+	new Vector3(-1.0, 0.7, 0.3),
+	new Vector3(0.2, 1.2, -0.9)
+];
+
+let activeScene: Scene | undefined;
+let flowMaterial: StandardMaterial | undefined;
+let updateRegistered = false;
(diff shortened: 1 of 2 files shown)


xtreemze:
Local certification for this experiment has moved to #97.

#97 starts at the exact #95 gameplay-parity head and reapplies this PR's child delta byte-for-byte: `energyFlowPrototype.ts` blob `d938ad23b718d4f9e2ee307a193e7df607df8480` and `hitEffect.ts` blob `ff39a0747576227634bef3f3eb29f042d573ef4c` both match this PR. The resulting #97 diff relative to #95 is exactly two files (+506/0 and +26/-7).

Keep #78 open as provenance while #97 remains valid. The infrequent local executor should **not** spend a separate checkout certifying this obsolete #36-based head; certify #95 first, then #97. If #97 is abandoned or its ancestry changes materially, return here for provenance/reassembly.

PR #96, opened 2026-09-03
Add deterministic tower dynamics, drilling, and maintenance contracts

xtreemze:
Refs #89, #82, #92
Related: #72, #91
Supersedes #94 for integration/certification; #94 remains the provenance record for the deployment/slew/drill parent slice.

## Why this PR is consolidated

The local executor runs infrequently. #96 originally stacked a maintenance lifecycle on #94, but the complete stack is cleanly based on current `master` and changes only nine files: four engine-free gameplay modules, four deterministic Storybook stories, and the Storybook README. There are no live Babylon/Cannon call-site changes, dependency changes, package changes, or production timing/balance constants.

Retargeting this top-of-stack PR to `master` means one exact head can certify the complete tower-contract family in one root compatibility pass and one Storybook install/build/browser session.

### Tower deployment

`src/js/gameplay/towerDeployment.ts`

- semantic stages: foundation / base / drilling / assembly / calibrating / ready / dry;
- caller-owned component windows;
- normalized growth progress;
- no baked production T1/T2/T3 timings.

### Turret dynamics

`src/js/gameplay/turretDynamics.ts`

- normalized yaw and shortest-angle delta;
- outward rest heading from silo→tower geometry;
- persistent angular velocity;
- acceleration-limited slew and braking-distance deceleration;
- independent max angular speed/acceleration;
- aim tolerance plus a separate angular-speed firing gate.

### Geothermal drilling

`src/js/gameplay/geothermalDrill.ts`

- 3D distance and downward-only finite reach;
- nearest point-source search;
- nearest point on polyline stream segments;
- nearest reachable stream contact point;
- no autonomous retries: geology may move while a dry tower remains dry until an explicit later search.

### Maintenance / renovation

`src/js/gameplay/towerMaintenance.ts`

State machine:

`isolating → opening → redrilling → reconnecting → calibrating → complete`

Rules:

- firing remains locked for the entire service cycle;
- internal service state/drill extension is explicit;
- geological re-search occurs exactly once at the redrill boundary;
- migrated geology alone never mutates tower state;
- successful retry preserves connection identity/contact point and reconnection glow;
- calibration must finish before firing unlocks;
- a failed retry completes as serviced-but-still-dry rather than inventing an automatic retry.

The schedule remains caller-owned; no production maintenance duration/cost is committed here.

## Storybook playgrounds

The stack adds four deterministic `Foundations/Defense/*` stories:

- Tower Deployment;
- Tower Dynamics;
- Geothermal Drill;
[...]

Issue #92, opened 2026-09-03
Plan infrequent local certification campaigns and evidence handoff


xtreemze:
Parent: #30
Related: #31, #52, #64, #66

## Goal

The local executor runs infrequently and is the only executor with reliable access to a real checkout, dependency installation, browser/runtime interaction, screenshots/video and profiling. Treat each session as a **certification campaign**, not one PR at a time.

Maximize useful evidence per setup while keeping the concurrently used development checkout untouched.

## Durable execution model

Use a dedicated sibling certification clone. Never switch branches, stash, reset, clean, install dependencies, or run certification tooling in the shared development checkout.

```sh
git clone https://github.com/xtreemze/defend.git ../defend-certification
cd ../defend-certification
git fetch --all --prune
git checkout --detach <exact-sha>
```

Evidence applies only to the exact SHA tested. Refresh `master` and all campaign heads immediately before execution; if a head moved materially, inspect the delta and recertify runtime-sensitive behavior.

For every tested SHA record:

```text
repository SHA:
branch/PR:
date/time:
OS + architecture:
Node / npm / pnpm:
Rust / cargo / wasm-pack:
browser + exact version:
git status --short before:
```

For each command record command, exit code and material stdout/stderr. Record `git status --short` afterward. Every lane ends explicitly in `PASS`, `FAIL`, or `BLOCKED`.

Dependency caches/node_modules may be reused only when package metadata is identical; record reuse rather than allowing it to hide dependency differences.

---

# Current frozen preparation cut — 2026-09-04

Anything opened after this cut belongs to a later campaign unless explicitly folded into a rebuilt certification head. Do not keep moving these targets during a local session.

## State 1 — current `master`

Current SHA: `9c5eaa9d93a7037b55e858f4ea14f8eace67f2fa`.

#35 is merged, so root validation defaults are safe:

```sh
npm install --package-lock=false --no-audit --no-fund
npm ls --all --json > /tmp/defend-npm-tree.json
npm run lint
npm run build
```

`lint` must remain non-mutating and `build` must terminate. If the historical dependency graph cannot install/build, capture the first hard blocker and move to independent lanes rather than spending the rare session on broad modernization.

### Historical game baseline

[...]

Issue #89, opened 2026-09-03
Design tower deployment, geothermal drilling, turret slew, and deformable foundations


xtreemze:
Parent: #29
Related: #72, #82, #86, #75, #87, #88

## Purpose

Make tower construction, power acquisition, aiming, and terrain interaction physically legible rather than instantaneous abstractions.

### Placement and drilling
- Towers may be placed anywhere the surface can physically support their footprint; geothermal proximity is not a placement gate.
- T2/T3 deploy with a subsurface siphon/drill that searches downward and locally outward for a reachable teal magma stream.
- If the drill fails to find a source, the tower remains structurally valid but cannot fire.
- Renovation/maintenance triggers a new drilling attempt, allowing previously dry towers to become powered after geothermal streams migrate or re-emerge nearby.
- A successful connection remains bounded by conduit/drill reach and shares source depletion rules from #82.
- T1 has no firing-energy requirement but still participates in foundation/terrain stabilization.

### Staggered physical deployment
Tower construction is a readable sequence, not an instantaneous mesh spawn. Prototype sequence:
1. foundation contacts terrain and settles;
2. nearby ground is gently stabilized toward the local average elevation;
3. base unfolds/grows;
4. drill/siphon descends and probes;
5. pillar grows for T2/T3;
6. turret structure assembles;
7. exposed internals/energy path illuminate if a source is found;
8. turret performs a restrained initialization sweep and settles to its rest heading.

T1 stops after the base/foundation sequence. T2 and T3 should have visibly distinct timings/silhouettes so the player can identify the resulting tier before assembly finishes.

### Turret rest orientation and slew
- Idle turret heading points generally away from the silo rather than using an arbitrary global direction.
- Tracking has finite angular speed/acceleration; `lookAt()`-style instantaneous snapping is not acceptable for the intended design.
- T2 should acquire/track faster than T3; T3's larger assembly should feel heavier.
- A turret should fire only when aim error is within an experimental tolerance; this makes fast R1 bodies and abrupt target changes tactically different from slow heavy targets.
- Target prediction remains useful, but the barrel slews toward the predicted heading over time.

### Deformable surface
- The playable surface has subtle elevation rather than being perfectly flat.
- Tower/wall foundations stabilize a bounded patch around themselves, nudging the local terrain toward an average support height rather than flattening the entire area.
- Stabilization should preserve drainage channels and broad terrain identity; it is not a terraforming tool.
- Volcanic activity can raise/bulge/fracture terrain.
[...]

Issue #82, opened 2026-09-03
Design geothermal tower power, planetary mantle flow, and eruptions


xtreemze:
Parent: #29
Related: #72, #73, #75, #76, #79

## Purpose

Make tower projectile energy physically sourced from the same conserved teal material that exists in raiders, silos and motherships.

The defending world is a spherical planet with a stylized solid interior, deep teal energy-bearing molten mantle, thin crust/playable surface, and near-surface streams/pockets/vents.

## Canonical rules

- Tower 2/3 do not create free projectile energy.
- A firing tower must connect to a reachable geothermal source through a visible finite-length curved siphon/conduit.
- Tower 1 remains a passive barrier and does not require a firing feed.
- Conduit viability becomes part of firing-tower placement.
- Several towers may share one source.
- Sustained draw can locally deplete a source even though the deep mantle is much larger.
- Prolonged depletion can cause a near-surface stream to retreat; over longer time another access route may emerge elsewhere.
- Prolonged accumulation/under-use can raise local pressure and eventually cause a telegraphed eruption.
- Eruptions are faction-neutral physical hazards: they can damage/melt towers and damage/redirect/destroy raiders.
- Raider energy liberated during eruptions still spills/flows toward the silo under #76 rules.
- Eruption material itself is not automatically free silo income; ownership remains explicit.
- Internal planet layers may be visible with low opacity/LOD so surface gameplay remains dominant.

## Spatial strategy

The system should create meaningful placement constraints without becoming factory plumbing:

- rich source + many towers = powerful but vulnerable to shared depletion;
- distributed sources = lower concentration risk and more resilient firing coverage;
- exhausted sources can force relocation/rebuilding;
- volatile rich zones can be tactically valuable but dangerous;
- raiders can pressure sectors where defender fire depends on one overdrawn source.

## Eruption sequence

Prototype should telegraph:

1. rising concentration/pressure;
2. brighter cracks/subsurface flow;
3. ground bulge;
4. widening fractures;
5. physical surface displacement;
6. teal magma/energy expulsion;
7. faction-neutral collision/heat damage;
8. post-eruption depleted/retreated source state.

Avoid invisible random damage circles.

## Performance architecture

Authoritative simulation should use low-count reservoir/stream graph state rather than volumetric mantle fluid:

- reservoir energy;
- pressure;
- deep replenishment;
- connected tower draw;
- conduit route/length;
- retreat/migration state;
- eruption state.

[...]

PR #98, opened 2026-09-03
Add deterministic geothermal reservoir lifecycle

xtreemze:
Refs #82, #89, #92
Related: #87, #91, #96

## Purpose

Extract the scalar geothermal-source lifecycle currently embedded in the Babylon geothermal lab into one renderer/geometry-independent contract that can later be shared by tower power, AI, the spherical planet simulation, and deterministic tests.

This PR does **not** modify the existing `geothermalPrototype.ts` call sites yet and does not promote its current calibration values into production balance.

## `src/js/gameplay/geothermalReservoir.ts`

Adds a deterministic local reservoir state model with:

- phases: `active / erupting / retreating`;
- bounded accessible-energy capacity;
- deep replenishment;
- requested versus actually served tower draw;
- unmet-energy accounting;
- depletion threshold + persistence timer before retreat;
- pressure accumulation weighted by reservoir fullness and suppressed/relieved by sustained consumption;
- eruption threshold;
- eruption-duration energy loss;
- frame-rate-independent pressure relaxation during eruption;
- post-eruption transition toward retreat;
- retreat-duration readiness signal for world-owned relocation;
- explicit re-seeding helper after the world chooses a new emergence point.

The contract deliberately does **not** choose a random relocation position, create particles, melt structures, route conduits, assign towers, or decide terrain deformation. Those remain world/geometry responsibilities.

## Storybook playground

Adds `Foundations/Physics/Geothermal Reservoir`.

The deterministic fixture compares two consequences of one parameter set:

1. **concentrated sustained draw** — local accessible energy is exhausted, remains below the depletion threshold, then retreats and becomes ready for relocation;
2. **rich unused source** — energy stays abundant, pressure accumulates, eruption begins, pressure/energy discharge, then the source transitions toward retreat.

The chart overlays accessible energy and pressure and marks retreat / relocation / eruption events.

Controls expose simulation duration, heavy draw rate, deep replenishment and pressure build rate.

The Storybook `play` test additionally simulates at 30 Hz and 60 Hz and checks retreat/eruption event timing stays within 0.2 s when both events occur, protecting the lifecycle against render-cadence dependence.

## Calibration status

The default fixture mirrors the existing geothermal PoC scale where useful (for example 0.075 depleted threshold, 6 s depletion persistence, 3.2 s retreat, 0.0055/s deep replenish, 0.018/s pressure build, 0.96 eruption threshold, 5.5 s eruption). These values remain **experimental calibration**, not committed campaign balance.

## Scope

- 3 commits ahead / 0 behind;
- 2 added files;
- no Babylon/Cannon/production call-site changes;
- no random number generation;
- no new background loop or persistent Storybook resource.

## Certification

Include this head in the next #92 root + Storybook campaign:

- safe root TypeScript/build compatibility for `geothermalReservoir.ts`;
- from `experiments/storybook`: `pnpm typecheck`, `pnpm build`, `pnpm test`;
- exercise draw, replenish, pressure-build and duration controls;
- verify sustained draw reaches retreat while idle-rich conditions can reach eruption;
- compare 30/60 Hz event timing;
- inspect all values for finite/bounded behavior at control extremes.

## Follow-up

After certification, update the already-merged `geothermalPrototype.ts` to consume this reservoir contract, deleting its local `energy / pressure / depletedSeconds / retreatSeconds / eruptionSeconds` transition math. A later integration can then combine this with #96's drilling/maintenance contracts without coupling scalar reservoir evolution to the temporary flat/mesh representation.

### DIFF
--- a/experiments/storybook/src/Foundations/Physics/GeothermalReservoir.stories.ts
+++ b/experiments/storybook/src/Foundations/Physics/GeothermalReservoir.stories.ts
@@ -1,0 +1,219 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	createGeothermalReservoirState,
+	stepGeothermalReservoir,
+	type GeothermalReservoirConfig,
+	type GeothermalReservoirPhase,
+	type GeothermalReservoirState
+} from "@defend/gameplay/geothermalReservoir";
+import { createLabShell } from "../../labTheme";
+
+type ReservoirArgs = {
+	simulationSeconds: number;
+	heavyDrawPerSecond: number;
+	deepReplenishPerSecond: number;
+	pressureBuildPerSecond: number;
+};
+
+interface TracePoint {
+	time: number;
+	energy: number;
+	pressure: number;
+	phase: GeothermalReservoirPhase;
+}
+
+interface ScenarioTrace {
+	points: TracePoint[];
+	finalState: GeothermalReservoirState;
+	eruptionAt: number | null;
+	retreatAt: number | null;
+	relocationAt: number | null;
+}
+
+function config(args: ReservoirArgs): GeothermalReservoirConfig {
+	return {
+		capacity: 1,
+		depletedThreshold: 0.075,
+		retreatDelaySeconds: 6,
+		retreatDurationSeconds: 3.2,
+		deepReplenishPerSecond: args.deepReplenishPerSecond,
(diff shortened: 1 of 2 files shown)


xtreemze:

Both #98 files were copied into #105 with matching blob SHAs, so the geothermal-reservoir behavior can be certified in the same root TypeScript and Storybook session as the other deterministic contracts. Keep this PR open as provenance; post any reservoir-specific failure here and rebuild only that slice of the batch if needed.

PR #95, opened 2026-09-03
certification: batch behavior-preserving gameplay seams

xtreemze:
Refs #30, #32, #92
Supersedes #36, #43, #45, #46 and #51 **for local certification only**. Keep those original PRs open as provenance until this batch is certified/merged or abandoned.

## Why this batch exists

The local executor runs infrequently. The five remaining behavior-preserving gameplay PRs are mechanically independent but each otherwise requires a separate checkout/build/browser session. This PR assembles their exact reviewed file blobs on top of current `master` so one exact SHA can certify all five ownership boundaries and their interactions in one runtime campaign.

This is intentionally a certification/integration candidate, not a new design change.

## Base and provenance

Base at assembly: `master` `9c5eaa9d93a7037b55e858f4ea14f8eace67f2fa`.

Source heads:

- #36 economy parity: `557dba7285d4d5152f3837db9ad11db2f0f295ba`;
- #43 tower metadata: `4040113d45e8070e36c81dcaee088fc9d400df27`;
- #45 projectile/turret scaling: `01beca0489f4c68a671794d36ab55bd6c43a5b8e`;
- #46 enemy scaling: `afc93016c70d64491729934b3dbef096bbb75a4b`;
- #51 enemy lifecycle/health meter: `22719a31fdaa156862dc4a3defc274da4fe5a6d9`.

Before assembly, GitHub commit comparisons confirmed that none of the 15 target files had independently changed on `master` since the corresponding source PR baseline. The five PR file sets are mutually disjoint. The batch therefore uses the exact source-branch blobs without hand-written conflict resolution.

### Economy — from #36

- `src/js/gameplay/economy.ts`;
- projectile hit recovery routed through the pure recovery/cap formula;
- bank collision damage routed through remaining-enemy-HP / 2.

No collision registration or balance change is intended.

### Tower identity — from #43

- explicit `{ level }` metadata on tower bases;
- pointer upgrade reads metadata rather than parsing a character from the mesh name.

Upgrade cost and current level-3 behavior are intentionally preserved for characterization.

### Projectile/turret scaling — from #45

- pure projectile HP / impulse / mass / shot-interval helpers;
- live call sites route through those helpers with unchanged level²/level³ arithmetic.

### Enemy scaling — from #46

- pure enemy diameter / HP / mass / movement impulse / restitution helpers;
- live spawn and movement call sites use those helpers;
- current negative restitution values remain intentionally preserved for Cannon characterization.

### Enemy finite lifetime — from #51

[...]

Issue #63, opened 2026-09-03
Build a spatial audio fly-by, occlusion, and visualization lab


xtreemze:
Parent: #60
Related: #52, #53, #56, #58, #61, #62

## Purpose

Create a controlled browser fixture for decisions that should be measured/listened to rather than guessed in production: moving-source Doppler, HRTF/equal-power cost, transform update cadence, voice virtualization, occlusion/transmission, near-field behavior, and restrained sound-wave visualization.

This is a certification/design lab, not a production integration PR.

### Web Audio transform/rendering model

The Web Audio 1.1 Editor's Draft models source/listener position and orientation through `PannerNode` and `AudioListener`, and AudioWorklet processes block/render-quantum audio on the rendering thread. Position values are `AudioParam`s and may be scheduled; do not stream one message per audio block from gameplay code.

Sources:
- https://webaudio.github.io/web-audio-api/
- https://developer.mozilla.org/en-US/docs/Web/API/PannerNode/positionX
- https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletProcessor/process

### Quality choices

`PannerNode` exposes `equalpower` and HRTF spatialization. The lab should measure actual CPU/glitch behavior with representative voice counts before #62 tiers map to concrete panning quality.

Source:
- https://developer.mozilla.org/en-US/docs/Web/API/PannerNode/panningModel

### Prior art for richer spatial semantics

- `bevy_sonus` (Bevy 0.19) is useful reference code for multi-ray occlusion, 3-band material transmission, diffraction, distance attenuation and lock-free ECS→DSP parameters. It is new/small; reference only, not dependency commitment: https://github.com/zem-invictus/bevy_sonus
- Resonance Audio demonstrates useful prior concepts including directivity, occlusion intensity, near-field gain, quality tiers and HRTF/ambisonic rendering, but its web SDK was archived on 19 Apr 2026 and must not become a new dependency: https://github.com/resonance-audio/resonance-audio-web-sdk

### A. Projectile fly-by

Stationary listener/camera. Launch identical procedural projectile emitters past it at multiple:
- velocities;
- miss distances;
- heights;
- approach angles;
- projectile tiers/masses.

Capture predicted closest approach from #61 and compare with the audible event.

### B. Moving listener

Orbit/pan the listener around stationary and moving emitters. Derive listener velocity from timestamped camera transforms with teleport/reset detection. A camera reset must not create an extreme Doppler transient.

### C. Voice-pressure stress

[...]

PR #99, opened 2026-09-03
Add bounded spatial motion estimation primitives

xtreemze:
Refs #60, #63
Related: #61, #62, #92

## Purpose

Add the missing timestamped motion layer required by the spatial-audio fly-by lab before any browser/Web Audio bridge starts deriving listener/source velocity ad hoc, together with a deterministic Storybook playground for inspecting the contract visually.

This PR remains isolated from production sound/gameplay. It does not touch WAFXR/Tone, AudioWorklet, Worker code, Babylon camera wiring, production sound routing, or gameplay.

## Pure motion contract

`src/js/audio/spatialMotion.ts`

Adds:

- `SpatialMotionSample` — timestamped world-space position;
- `SpatialMotionCalibration` — minimum/maximum sample delta, teleport threshold, velocity smoothing half-life, maximum speed and bounded prediction horizon;
- `SpatialMotionEstimate` — sanitized position, smoothed bounded velocity, timestamp, observed sample delta and explicit reset flag;
- `initialSpatialMotionEstimate()`;
- `updateSpatialMotionEstimate()`;
- `spatialMotionSpeed()`;
- bounded short-horizon position prediction helpers for lower-rate control snapshots.

## Reset semantics

Camera resets, scene loads, debugger scrubs and other discontinuities must not become artificial Doppler spikes.

The estimator resets velocity to zero when:

- timestamps are non-monotonic;
- the sample gap exceeds the configured maximum, when `maxSampleDeltaSeconds > 0`;
- the position jump meets/exceeds the configured teleport distance, when `teleportDistance > 0`.

`maxSampleDeltaSeconds <= 0` explicitly disables long-gap reset. `teleportDistance <= 0` explicitly disables distance-based reset.

Very small positive sample deltas do not reset. Instead, the velocity denominator is floored by `minSampleDeltaSeconds`, preventing transform jitter from exploding into extreme apparent speed.

## Smoothing

Velocity smoothing is expressed as a time-domain half-life, not a per-frame lerp coefficient. This keeps behavior substantially independent of render/display cadence.

`velocityHalfLifeSeconds <= 0` selects the current bounded raw velocity directly.

`maxSpeed` clamps both previous and newly derived velocities. A non-positive maximum produces zero velocity rather than non-finite output.

## Prediction boundary

`predictSpatialMotionPosition()` only performs bounded forward extrapolation from an already-derived estimate. It is intended for timestamped lower-rate audio/control snapshots and interpolation experiments in #63.

It is explicitly not simulation authority and must never advance gameplay state.

## Numeric boundary

The module mirrors #61's defensive finite-state philosophy:

- NaN maps to neutral/fallback values;
- extreme/infinite scalar state is saturated at an internal defensive ceiling;
- vector math and speed outputs remain finite;
- malformed timing resets instead of poisoning downstream Doppler/closest-approach calculations.

This is robustness, not world/balance calibration.

## Storybook playground

Adds `Audio/Spatial/Motion Estimator` in:

`experiments/storybook/src/Audio/Spatial/SpatialMotion.stories.ts`

The story is deterministic and starts no RAF, AudioContext, Worker or timer. Controls expose:

- sample cadence;
- source/listener speed;
- smoothing half-life;
- maximum derived speed;
- teleport threshold;
- teleport on/off.

The default fixture runs a 1.5-second timestamped trajectory with one persistent position discontinuity. It visualizes:

- the sampled world-space path;
- the discontinuity/reset marker;
- final smoothed velocity vector;
- bounded 120 ms forward prediction;
- reset count and final speed/limit metrics.

The story's interaction assertion expects exactly one runtime reset in the default teleport fixture and verifies the final speed remains finite and within `maxSpeed`.

## Why before the browser lab

#63 already requires a moving-listener fixture with timestamped camera transforms and teleport/reset detection. Centralizing that behavior now prevents the later Babylon camera adapter, Worker planner and Web Audio renderer from each inventing subtly different velocity semantics.

It also supports the planned transform-cadence comparison:

1. render-frame updates;
2. lower-rate timestamped snapshots;
3. Worker-planned snapshots;
4. optional SAB only if profiling justifies it.

## Scope check

- 3 commits ahead / 0 behind;
- exactly 2 added files;
- one pure root audio-math module;
- one deterministic Storybook story;
- no package/dependency/live-runtime changes.

This branch is independent of #95/#97/#96 and must not block that certification sequence.

## Local certification

Keep draft until the local campaign can cover both the root historical TypeScript target and Storybook gate.

### Root numeric/compatibility checks

1. TypeScript 3.2 accepts the module/import from `spatialAudio.ts`.
2. Stationary repeated samples remain exactly zero velocity.
3. Constant-velocity trajectories converge to the expected vector within smoothing tolerance across 30/60/120 Hz sample cadences.
4. A large teleport resets velocity to exactly zero and marks `reset=true`.
5. Non-monotonic timestamp resets.
6. Long sample gap above enabled `maxSampleDeltaSeconds` resets; non-positive maximum disables that reset.
7. Very small positive sample delta remains finite and speed stays <= `maxSpeed`.
8. NaN/Infinity/extreme finite position, time and calibration inputs never produce non-finite outputs.
9. Bounded prediction never extrapolates past `maxPredictionSeconds` and never predicts backward for past target timestamps.
10. Feed derived listener/source velocities into #61 Doppler/closest-approach fixtures and confirm camera reset produces no transient extreme ratio.

### Storybook gate

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Exercise `Audio/Spatial/Motion Estimator` across 30/60/120 Hz, smoothing extremes, speed-limit changes and teleport on/off. Confirm reset markers/metrics agree with the pure fixtures and the story leaves no persistent browser resource behind.

## Next

After this contract is certified, #63 should use it in the moving-listener/fly-by fixture. Only then should a Web Audio compatibility adapter decide transform scheduling cadence and modern AudioParam vs legacy setter behavior.

### DIFF
--- a/experiments/storybook/src/Audio/Spatial/SpatialMotion.stories.ts
+++ b/experiments/storybook/src/Audio/Spatial/SpatialMotion.stories.ts
@@ -1,0 +1,211 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+  initialSpatialMotionEstimate,
+  predictSpatialMotionFromCalibration,
+  spatialMotionSpeed,
+  updateSpatialMotionEstimate,
+  type SpatialMotionCalibration,
+  type SpatialMotionEstimate
+} from "@defend/audio/spatialMotion";
+import { createLabShell } from "../../labTheme";
+
+type MotionArgs = {
+  sampleHz: number;
+  speed: number;
+  halfLife: number;
+  maxSpeed: number;
+  teleportDistance: number;
+  includeTeleport: boolean;
+};
+
+type TracePoint = {
+  x: number;
+  z: number;
+  estimate: SpatialMotionEstimate;
+};
+
+type MotionTrace = {
+  points: TracePoint[];
+  resetCount: number;
+  finalSpeed: number;
+  predictedX: number;
+  predictedZ: number;
+};
+
+function buildTrace(args: MotionArgs): MotionTrace {
+  const sampleHz = Math.max(1, Math.floor(args.sampleHz));
+  const deltaSeconds = 1 / sampleHz;
+  const calibration: SpatialMotionCalibration = {
+    minSampleDeltaSeconds: 1 / 500,
+    maxSampleDeltaSeconds: 0.25,
+    teleportDistance: Math.max(0, args.teleportDistance),
+    velocityHalfLifeSeconds: Math.max(0, args.halfLife),
+    maxSpeed: Math.max(0, args.maxSpeed),
+    maxPredictionSeconds: 0.12
+  };
+  const startX = -72;
+  const startZ = -22;
+  const duration = 1.5;
+  let offsetX = 0;
+  let offsetZ = 0;
+  let teleported = false;
+  let estimate = initialSpatialMotionEstimate({
+    position: { x: startX, y: 0, z: startZ },
+    timeSeconds: 0
+  });
+  const points: TracePoint[] = [{ x: startX, z: startZ, estimate }];
+  let resetCount = 0;
+
+  for (let timeSeconds = deltaSeconds; timeSeconds <= duration + 1e-9; timeSeconds += deltaSeconds) {
+    if (args.includeTeleport && !teleported && timeSeconds >= 0.75) {
+      const distance = Math.max(1, calibration.teleportDistance);
+      offsetX = distance + 28;
+      offsetZ = -(distance * 0.45 + 10);
+      teleported = true;
+    }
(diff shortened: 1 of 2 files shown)


xtreemze:

Both #99 files are present in #105 with matching blob SHAs. The motion estimator can therefore use the same root TypeScript and Storybook campaign as the other deterministic contracts. Keep this PR open as provenance; post any motion/Doppler-specific failure here and rebuild only this slice if necessary.

Issue #83, opened 2026-09-03
Design three-set raid planning, inertial mothership movement, and silo exclusion


xtreemze:
Parent: #79
Related: #73, #74, #75, #80, #81, #82

## Purpose

Define the raider-side planning and movement layer so the mothership remains a finite physical vessel rather than an RTS camera.

## Canonical direction

- Player/AI plans the upcoming three raid sets.
- Each set can contain an R1/R2/R3 composition and a general target sector.
- A set may intentionally be left empty/held; no raid is required at a fixed cadence.
- Waiting is not free because mothership hover consumes the same teal reserve.
- Input selects a desired mothership/target position; the ship accelerates toward it with mass, inertia, persistent velocity and damping.
- Camera follows the actual moving mothership with comfort-oriented damping; it does not teleport independently.
- Movement consumes a small amount of the same reserve, not a separate fuel currency.
- The mothership cannot safely hover above the silo.

## Silo attraction / exclusion

The silo is rooted into the planet and coupled to a much larger grounded energy system. Because teal energy tends to pool/coalesce, an exposed mobile mothership reserve entering the silo's strongest field is physically pulled toward it.

Implement this as a continuous attraction/instability field, not an invisible kill cylinder:

- warning region raises stabilization cost and bends trajectory;
- critical region can overpower propulsion and capture/pull down the mothership;
- target-position solver normally projects invalid desired positions outside the unsafe region;
- inertia can still carry the ship into danger;
- exact radii/force curves remain experimental.

## Target sectors

Tapping a surface sector selects the general drop area for an upcoming raid set. It does not teleport raiders onto the ground.

Actual insertion should derive from mothership position/velocity, selected sector, raider type, launch impulse, surface curvature, obstacles and bounded dispersion.

The mothership can therefore attack the same sector from different offset positions/angles without ever hovering directly over the silo.

## Dynamic cadence

The three-set horizon supports probe/feint/follow-through patterns and deliberate holds. AI should use the same high-level constraints: finite reserve, expected extraction, visible tower/geothermal state, prior outcomes and target-sector choice.

## Geothermal interaction

Raid planning should eventually consider visible defender geology:

- weakly supplied tower sector;
- source shared by too many towers;
- source near depletion/retreat;
- impending eruption;
- route likely to force sustained fire and exhaust local supply.

## PoC extension for #81

Extend the mothership lab in small steps:

1. tap/click desired lateral target;
2. bounded acceleration + persistent velocity + damping;
[...]

Issue #74, opened 2026-09-03
Prototype the symmetric raid economy, deterrence model, and AI-defended inversion


xtreemze:
Parent: #73
Related: #30, #32, #50, #72

## Purpose

Turn the campaign inversion into measurable mechanics before changing production balance. This issue owns experimental formulas/fixtures for the **defender → deterrence → mothership → raider** transition. It should not directly rebalance production constants until deterministic/browser evidence exists.

## 1. Starting fortress hypothesis

Start with a minimal functional fortress that already contains all three tower roles.

A particularly coherent baseline is exactly **one Tower 1, one Tower 2, and one Tower 3** around the silo.

At current cumulative construction economics:

- T1 replacement value: 3000;
- T2 replacement value from an empty cell: 3000 + 6000 = 9000;
- T3 replacement value from an empty cell: 3000 + 6000 + 9000 = 18000;
- total replacement value of one-of-each = **30000**, exactly the current maximum silo balance.

This is narratively useful: the player begins with a complete but minimal inherited defensive system whose full replacement value equals the entire maximum reserve. Losing infrastructure matters immediately, and the player cannot casually duplicate the whole fortress while also keeping a safety reserve.

Do not assume this is final. Compare one-of-each against variants with additional Tower 1 barriers, but preserve the principle that the starting fortress is functional, sparse, and expensive enough to make preservation meaningful.

## 2. Symmetric raider launch-cost hypothesis

The current defensive collision rule removes approximately `remaining enemy HP / 2` from the silo.

Treat the same quantity as candidate **energy extracted by a raider** after the role reversal.

Current full-health extraction potential:

| Raider tier | HP | max `HP / 2` extraction |
| --- | ---: | ---: |
| 1 | 15440 | 7720 |
| 2 | 60880 | 30440 |
| 3 | 136320 | 68160 |

First launch-cost hypothesis:

`launchCost(level) = 3000 × level²`

which gives:

| Raider tier | launch cost | full-health extraction | full-health gross return / cost |
| --- | ---: | ---: | ---: |
| 1 | 3000 | 7720 | 2.57× |
| 2 | 12000 | 30440 | 2.54× |
| 3 | 27000 | 68160 | 2.52× |

More importantly, break-even remaining HP is almost tier-independent:

`remaining HP at break-even = 2 × launchCost`

- R1: 6000 / 15440 = **38.9%** health;
- R2: 24000 / 60880 = **39.4%** health;
[...]

Issue #73, opened 2026-09-03
Campaign arc: fortress defense, deterrence collapse, and the mothership raider inversion


xtreemze:
Parent: #29
Related: #30, #32, #50, #72

## Core premise

The campaign should begin from a position of **responsibility rather than expansion**. The player inherits a small but functional fortress protecting a finite energy silo. Defense and survival are paramount; resource accumulation is secondary and exists to sustain that defense.

The fortress begins **prebuilt and minimal**, already demonstrating all three tower roles:

- **Tower 1 / Barrier** — spatial control and delay;
- **Tower 2 / Interceptor** — frequent, granular defensive fire;
- **Tower 3 / Siege Cannon** — long-range heavy impulse.

This avoids starting from an empty sandbox and immediately establishes that the player is the custodian of something valuable, not an abstract builder placing pieces onto a board.

## Narrative inversion

The defended silo has a structural dependency: raids are dangerous, but they are also one of the only meaningful ways to replenish energy.

The player initially understands raiders only as a threat. Successful defensive combat recovers energy and keeps the fortress alive. As the fortress becomes increasingly effective, attackers begin to judge the silo as too costly to raid. Raid frequency declines. Eventually the attacks stop entirely.

This should initially feel like victory.

It is not.

Without incoming raids there is no meaningful energy inflow. Towers continue to age and fall into disrepair. The fortress consumes its remaining reserve maintaining the silo and surviving. The player discovers that **perfect deterrence has made the defended system unsustainable**.

At a critical energy threshold, the silo uses its last strategic reserve to enact a survival protocol: it converts itself into a mobile mothership and abandons the fortress.

The player who spent the first act defending against raiders is forced to become one.

## Why the inversion works mechanically

This should not be a detached story cutscene. The same simulation rules should change meaning when viewed from the other side.

### Defensive phase

- energy is health, construction budget, and defended resource;
- incoming raiders threaten the silo;
- projectile hits recover energy;
- towers age and require continuing resource flow;
- surviving an attack is good, but active combat is economically useful;
- over-investing in static defense can eventually suppress the very pressure that sustains the system.

### Raider phase

- the mothership carries a finite energy reserve;
- launching a raiding body costs energy;
- deployed raiders remain finite-lived and decay over time;
- a raider that reaches an enemy silo extracts/steals energy based on its remaining viability;
- destroyed, ejected, delayed-until-decay, or badly damaged raiders generate little or no return;
- successful raids must be net-positive or the mothership eventually dies.

[...]

PR #100, opened 2026-09-03
Add deterministic three-set raid planning contract

xtreemze:
Refs #83, #92
Related: #73, #74, #87

## Purpose

Extract the reusable three-set raid-planning semantics from the merged mothership/navigation prototype without carrying over DOM, keyboard listeners, animation-frame refresh, debug globals, or temporary launch-cost assumptions.

This PR does **not** implement raider spawning, mothership movement, launch energy spend, AI planning, or final sector geometry.

## `src/js/gameplay/raidPlanning.ts`

Adds an engine-free immutable planning contract:

- exactly three upcoming raid sets;
- R1/R2/R3 integer counts with caller-supplied per-tier cap;
- missing sets normalize to intentional HOLD;
- extra sets are discarded so the planning horizon cannot silently grow;
- immutable tier-count set/adjust helpers;
- explicit HOLD/clear operation;
- optional per-set sector snapshot;
- deterministic queue advance (`2 → 1`, `3 → 2`, fresh HOLD → `3`);
- body-count summary;
- caller-supplied commitment weights;
- launch-time sector resolution: use a saved snapshot when present, otherwise resolve against the current ground sector at launch;
- copied sector values so later UI/navigation mutation cannot retroactively change a committed target.

The gameplay contract intentionally owns **queue semantics only**. The temporary 1/4/9 commitment weights from the hybrid lab are not baked into the module and do not become production launch costs.

## Storybook playground

Adds `Foundations/Topology/Raid Planning`.

The fixture shows:

- the full upcoming three-set horizon;
- a first swarm set whose sector is deliberately deferred until launch;
- a second mixed set whose sector is snapshotted in advance;
- an intentional HOLD third set;
- the queue after the first set launches;
- side-by-side resolution of deferred versus snapshotted sectors;
- lab-only commitment weights supplied by the story rather than the gameplay module.

Controls expose per-tier cap, representative R1/R2 composition, current sector, and snapshotted sector.

The Storybook assertion verifies:

- both pre/post queues contain exactly three sets;
- deferred sector resolution follows the current launch sector;
- snapshotted sector resolution remains fixed;
- queue advance appends a HOLD third slot.

## Scope

- 3 commits ahead / 0 behind;
- 2 added files;
- no balance-cost constants;
- no rendering or navigation changes.

## Certification

Include in the next #92 root + Storybook campaign when practical:

- root TypeScript/build compatibility for `raidPlanning.ts`;
- Storybook `pnpm typecheck`, `pnpm build`, `pnpm test`;
- exercise count caps and both sector controls;
- verify queue length is always three;
- verify HOLD survives normalization/advance;
- verify current-sector mutation does not alter a snapshotted target.

## Follow-up

After certification, replace the queue mutation logic inside the merged `experiments/hybrid-engine/src/raidPlanPrototype.ts` with this contract while leaving its lab UI presentation intact. Mothership launch-cost and reserve-spend rules should be layered later from explicit economy design rather than inferred from the old 1/4/9 display weights.

### DIFF
--- a/experiments/storybook/src/Foundations/Topology/RaidPlanning.stories.ts
+++ b/experiments/storybook/src/Foundations/Topology/RaidPlanning.stories.ts
@@ -1,0 +1,162 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	advanceRaidPlan,
+	assignRaidSector,
+	createRaidPlan,
+	raidSetCommitment,
+	raidSetIsHold,
+	resolveRaidSector,
+	type RaidCommitmentWeights,
+	type RaidSectorTarget,
+	type RaidSetPlan
+} from "@defend/gameplay/raidPlanning";
+import { createLabShell } from "../../labTheme";
+
+type RaidPlanningArgs = {
+	maximumPerTier: number;
+	currentSectorX: number;
+	currentSectorZ: number;
+	snapshotSectorX: number;
+	snapshotSectorZ: number;
+	set1R1: number;
+	set2R1: number;
+	set2R2: number;
+};
+
+const LAB_WEIGHTS: RaidCommitmentWeights = { r1: 1, r2: 4, r3: 9 };
+
+function composition(plan: RaidSetPlan): string {
+	if (raidSetIsHold(plan)) return "HOLD / NO RAID";
+	const parts: string[] = [];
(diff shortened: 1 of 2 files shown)


xtreemze:

Both #100 files are present in #105 with matching blob SHAs. Queue length/HOLD/sector-resolution behavior can therefore be certified in the same root TypeScript and Storybook session. Keep this PR open as provenance; post any raid-planning-specific failure here and rebuild only this slice if needed.

Issue #79, opened 2026-09-03
Define the raider mothership as a finite-energy physical vessel with crash defeat


xtreemze:
Parent: #73
Related: #29, #72, #74, #75, #76, #77, #78

## Purpose

Define the raider-side mothership as a first-class physical system that obeys the same visual, economic, and simulation principles as the rest of Defend. It must not become a conventional sci-fi gunship, menu hub, or invulnerable camera platform.

The mothership is the mobile lifecycle state of a silo: a finite-energy transport, launch platform, observer, extraction vessel, and eventually a physical wreck if the raid fails.

## Canonical visual rule

**Defense is rectilinear and rooted; raiding is spherical and mobile.**

### Defensive family

- silo / fortress geometry is broad, low, blocky, grid-aligned and anchored to the ground;
- Tower 1/2/3 use rectilinear bases, pillars and turret structures;
- defensive construction reads as architecture and obstruction.

### Raider family

- raiders are spherical / rounded low-poly bodies;
- the mothership should look like a larger relative of those raiders rather than a flying tower base;
- use simple generated geometry, visible structure and exposed internals;
- avoid smooth authored spaceship hulls, wings, cockpits, panel-detail realism, decals, or conventional military spacecraft language.

## Mothership form

The mothership should be organized around a **central, visibly finite teal energy reservoir**.

Suggested geometric language:

- a large central teal core / liquid reservoir;
- an incomplete geodesic or faceted spherical outer shell;
- gaps in the shell deliberately expose the core and machinery;
- simple radial struts connect shell sections to the core;
- one or more open rings / cages define raider storage, launch and recovery paths;
- small spherical or toroidal thruster/field elements sit around the outer shell rather than conventional engines;
- asymmetry from missing/degraded shell segments is acceptable and can communicate damage or low energy;
- no opaque hull should fully conceal the energy state.

The transformed silo should remain recognizable in the mothership through the teal core and conserved energy volume even though the surrounding blocky ground architecture has reconfigured into the spherical raider visual family.

## The mothership has no weapon

The mothership should not shoot.

It exists to:

- carry finite energy;
- remain aloft;
- observe the target;
- create/deploy raider bodies;
- recover surviving raiders;
- siphon extracted energy;
- choose whether to continue raiding or settle/defend again.

The raiders themselves are the offensive payload.

## Raider weapon principle

**Raiders do not carry weapons; their shell/body is the weapon.**

Their offensive effects come from:

- mass;
- momentum;
- collision;
[...]

PR #101, opened 2026-09-03
Add deterministic mothership navigation and silo-attraction contract

xtreemze:
Refs #83, #79, #92
Related: #87, #100, #102

## Purpose

Extract the reusable **horizontal** mothership-motion rules from the merged navigation lab so player input, AI, camera presentation, and later production physics can share one deterministic notion of inertia and silo attraction.

This PR intentionally does **not** own mothership reserve, hover, loss of lift, or crash. #102 is the proposed single reserve/lift authority. Navigation only reports propulsion-energy demand.

## `src/js/gameplay/mothershipNavigation.ts`

Adds an engine-free 2D navigation contract with:

- persistent position/velocity;
- desired ground raid sector distinct from desired mothership position;
- outward launch offset;
- safe-radius target projection when ordinary input would command a position too close to the silo;
- bounded engine steering acceleration;
- bounded arrival braking using the same engine acceleration ceiling;
- configurable cruise/approach speed and exponential damping;
- stable / warning / critical / captured phases from silo distance;
- continuous nonlinear inward silo-attraction field;
- capture only when the vessel is deep inside the critical region **and already moving inward above a threshold**;
- captured state removes steering authority and applies stronger inward attraction;
- horizontal overspeed cap;
- per-step propulsion-energy demand derived from **engine steering effort only**;
- cumulative movement-demand diagnostics;
- projected target count and capture-transition diagnostics.

### Intentional corrections from the lab prototype

The merged `mothershipNavigationPrototype.ts` contains three shortcuts that should not become shared gameplay semantics:

1. its near-target braking (`velocity * -1.2`) can exceed nominal max engine acceleration; the extracted contract bounds braking to engine capability;
2. its movement-energy accounting charges steering **plus external silo attraction**; the extracted contract reports propulsion demand from engine steering only;
3. it directly decrements a local reserve and hover drain. The extracted contract owns no reserve at all; #102 is intended to apply movement demand together with hover, launch spend, and extraction inflow to one conserved teal pool.

These are physical/ownership corrections, not balance tuning.

## Storybook playground

Adds `Foundations/Physics/Mothership Navigation`.

The deterministic fixture overlays:

- safe / warning / critical silo radii;
- a near-silo ground sector whose desired ship position is projected outward to the safe radius;
- a commanded far reposition trace showing acceleration/inertia/braking;
- a deliberately dangerous inertial approach starting inside the critical region with inward velocity, demonstrating that safe target projection does not erase momentum risk;
- propulsion-demand and steering diagnostics;
- 30 Hz versus 60 Hz capture timing.

The default capture fixture preserves the existing PoC condition shape: capture radius factor 0.55 and inward-speed threshold 3, while all calibration values remain Storybook/PoC values rather than production campaign balance.

The Storybook assertion verifies:

- the near-silo target is projected;
- projected desired position is outside the 35-unit lab safe radius;
- engine steering never exceeds configured max acceleration;
- when capture occurs at both 30 Hz and 60 Hz, event timing differs by less than 0.08 s.

## Scope

Current head: `adb584b6d68e23f3bc3fed07e2c4331f4c60c93f`.

Relative to `master` at branch creation:

- 2 files only;
- no live Babylon/Cannon call-site changes;
- no camera implementation changes;
- no vertical crash/lift changes;
- no reserve authority;
- no final balance constants.

## Certification

Include in the next #92 root + Storybook campaign when practical:

- root TypeScript/build compatibility for `mothershipNavigation.ts`;
- Storybook typecheck/build/test;
- exercise acceleration, damping, attraction and capture-threshold controls;
- verify normal sector targeting cannot directly command the ship inside the safe radius;
- verify inertia can nevertheless enter warning/critical/capture states;
- verify engine steering remains bounded during arrival braking;
- verify propulsion demand follows steering effort and becomes zero after captured steering authority is removed;
- compare 30/60 Hz traces and capture timing.

## Follow-up

After certification, update the merged `mothershipNavigationPrototype.ts` to consume this contract while leaving Babylon rendering/camera presentation intact. Its local reserve/drain logic should be replaced by #102 after that contract is certified. Separately, #100 can replace the prototype raid-plan queue; launch cost remains an explicit later economy decision.

### DIFF
--- a/experiments/storybook/src/Foundations/Physics/MothershipNavigation.stories.ts
+++ b/experiments/storybook/src/Foundations/Physics/MothershipNavigation.stories.ts
@@ -23,7 +23,6 @@ type NavigationArgs = {
 interface TracePoint {
 	time: number;
 	position: NavigationVector2;
-	reserve: number;
 	phase: string;
 }
 
@@ -54,7 +53,6 @@ function config(args: NavigationArgs): MothershipNavigationConfig {
 		captureInwardSpeed: args.captureInwardSpeed,
 		capturedAccelerationMultiplier: 1.35,
 		overspeedFactor: 1.35,
-		hoverDrainPerSecond: 0.0025,
 		movementDrainPerAcceleration: 0.00028
 	};
 }
@@ -74,7 +72,6 @@ function simulate(
 		points.push({
 			time: elapsed,
 			position: { x: state.position.x, z: state.position.z },
-			reserve: state.reserve,
 			phase: state.phase
 		});
 		const step = stepMothershipNavigation(state, deltaSeconds, calibration);
@@ -130,13 +127,12 @@ const meta = {
 		);
 		const safeTargetDistance = navigationVectorLength(projection.desiredPosition);
 
-		let farState = createMothershipNavigationState(start, 1, nearSector, calibration);
+		let farState = createMothershipNavigationState(start, nearSector, calibration);
 		farState = setMothershipRaidSector(farState, farSector, calibration);
 		const farTrace = simulate(farState, args.simulationSeconds, 1 / 60, calibration);
 
 		const captureBase = createMothershipNavigationState(
 			{ x: 13, z: 0 },
-			1,
 			nearSector,
 			calibration
 		);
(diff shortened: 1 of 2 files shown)


xtreemze:

Both #101 files are present in #105 with matching blob SHAs. Safe-target projection, bounded steering/braking, silo attraction, propulsion-demand separation and 30/60 Hz capture timing can therefore be certified in the same root TypeScript and Storybook session. Keep this PR open as provenance; post navigation-specific failures here and rebuild only this slice if required.

PR #102, opened 2026-09-03
Add authoritative mothership energy and loss-of-lift lifecycle

xtreemze:
Refs #79, #83, #92
Related: #87, #101, #100

## Purpose

Establish one authoritative mothership teal-energy pool and vertical lift/crash lifecycle, rather than letting navigation, launch planning, and the Babylon hover prototype each own their own reserve.

Other systems submit energy inflow/spend demand; this domain owns conservation, authorization semantics, suspension state, loss of lift, impact and hulk consequences. Category-level accounting remains with the caller for player-facing transparency.

## `src/js/gameplay/mothershipEnergyLift.ts`

Adds an engine-free lifecycle with:

- absolute configurable energy capacity rather than assuming the final game uses normalized 0–1 energy;
- stable / low-reserve / critical / falling / hulk phases;
- configurable low/critical/fall thresholds;
- continuous hover drain and separate falling drain;
- externally supplied energy inflow (extraction/recovery);
- externally requested spend (navigation, launch or other systems);
- applied versus unmet spend accounting;
- reserve-derived lift fraction and target-altitude sag;
- reserve-dependent lift stiffness/damping and deterministic instability;
- irreversible-by-default loss of lift, with explicit experimental `allowFallRecovery`;
- gravity-driven fall;
- bounded first impact/bounce and final persistent hulk.

## `src/js/gameplay/mothershipEnergyAuthorization.ts`

Makes the no-energy-debt rule executable before adapters are written:

- `authorizeDiscreteMothershipSpend()` is all-or-nothing for physical actions such as raider launch;
- it reports requested/available/shortfall and never mutates reserve;
- optional protected reserve allows later policy to preserve a survival/hover buffer without changing core physics;
- `fundContinuousMothershipSpend()` permits proportional funding for continuous propulsion/service demand rather than allowing implicit debt;
- `mothershipFundingFraction()` gives an actuator scaling factor;
- hulk has zero spend authority.

These helpers are intentionally non-mutating. The integration transaction is: authorize/fund → apply the permitted physical authority → submit exactly the funded spend to `stepMothershipEnergyLift()` as the single reserve mutation.

A discrete launch must not spawn first and discover afterward that the mothership could not pay.

## Reserve ownership boundary

#101 navigation owns **no reserve**. It reports propulsion demand only. #102 is intended to apply that demand together with hover drain, raid launch spend, and physically delivered extraction inflow.

External silo attraction is not propulsion and must continue to act even when engine funding falls to zero.

This prevents double-draining and keeps the exposed central teal core semantically tied to one conserved value.

## Storybook playground

Adds `Foundations/Physics/Mothership Energy & Lift`.

The fixture compares two otherwise identical raid horizons:

1. hover + external raid spending with no extraction;
2. the same spending with one successful mid-raid extraction pulse.

It overlays reserve and altitude and marks fall/hulk/extraction events, demonstrating that extraction extends the raid horizon without removing the continuous survival cost.

Default visual calibration mirrors the merged Babylon PoC where useful (capacity 1; low .38 / critical .20 / fall .10; hover drain .0065/s; falling drain .002/s; altitude 56→18; gravity 9.81; first-impact restitution .12). These remain experimental lab calibration, not final campaign balance.

The Storybook assertion verifies fall before hulk, representative extraction postponing fall, and close 30/60 Hz hulk timing.

## Scope

Current head: `1c363ec385fb65e7f53188bd8545f1e00dc4e537`.

Relative to branch base `master` `9c5eaa9d93a7037b55e858f4ea14f8eace67f2fa`:

- 3 commits;
- 3 added files;
- no live Babylon/Cannon call-site changes;
- no final reserve capacity, launch cost, or threshold commitment.

## Certification

Include in #92 when practical:

- root TypeScript/build compatibility for both gameplay modules;
- Storybook typecheck/build/test;
- reserve remains bounded by capacity;
- discrete authorization rejects a request larger than spendable reserve and hulk always rejects;
- continuous funding returns funded + unmet = requested and a bounded 0..1 funding fraction;
- no-extraction raid has a finite horizon;
- extraction extends that horizon under representative settings;
- 30/60 Hz fall/hulk timing;
- first-impact bounce and final persistent hulk.

## Follow-up

After certification, `mothershipPrototype.ts` should consume this reserve/lift contract and delete its local reserve/phase/vertical-velocity transition math. #101 navigation demand and later #100 launch demand should pass through the authorization/funding boundary before becoming physical actions.

### DIFF
--- a/experiments/storybook/src/Foundations/Physics/MothershipEnergyLift.stories.ts
+++ b/experiments/storybook/src/Foundations/Physics/MothershipEnergyLift.stories.ts
@@ -1,0 +1,287 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	createMothershipEnergyLiftState,
+	stepMothershipEnergyLift,
+	type MothershipEnergyLiftConfig,
+	type MothershipEnergyLiftState
+} from "@defend/gameplay/mothershipEnergyLift";
+import { createLabShell } from "../../labTheme";
+
+type LiftArgs = {
+	simulationSeconds: number;
+	externalSpendPerSecond: number;
+	extractionAtSeconds: number;
+	extractionEnergy: number;
+};
+
+interface TracePoint {
+	time: number;
+	reserve: number;
+	altitude: number;
+	phase: string;
+}
+
+interface LiftTrace {
+	points: TracePoint[];
+	finalState: MothershipEnergyLiftState;
+	fallAt: number | null;
+	hulkAt: number | null;
+}
+
+function config(): MothershipEnergyLiftConfig {
+	return {
+		capacity: 1,
+		lowReserve: 0.38,
+		criticalReserve: 0.2,
+		fallReserve: 0.1,
+		hoverDrainPerSecond: 0.0065,
+		fallDrainPerSecond: 0.002,
+		startAltitude: 56,
+		hulkCenterAltitude: 18,
+		reserveSagDistance: 12,
+		maxLiftStiffness: 4.4,
+		minimumLiftStiffnessFactor: 0.32,
+		baseVerticalDamping: 1.1,
+		additionalVerticalDamping: 3.1,
+		baseWobbleAmplitude: 0.08,
+		depletionWobbleAmplitude: 1.5,
+		baseWobbleFrequency: 1.2,
+		depletionWobbleFrequency: 1.8,
+		gravity: 9.81,
+		impactBounceThreshold: 4,
+		impactRestitution: 0.12,
+		allowFallRecovery: false
+	};
+}
+
+function simulate(
+	seconds: number,
+	deltaSeconds: number,
+	externalSpendPerSecond: number,
+	extractionAtSeconds: number | null,
+	extractionEnergy: number
+): LiftTrace {
+	const calibration = config();
+	let state = createMothershipEnergyLiftState(calibration.capacity, calibration);
+	const points: TracePoint[] = [];
+	let fallAt: number | null = null;
+	let hulkAt: number | null = null;
+	let elapsed = 0;
+	let extractionApplied = false;
+
+	while (elapsed <= seconds + 1e-9) {
+		points.push({
+			time: elapsed,
+			reserve: state.reserve / calibration.capacity,
+			altitude: state.altitude,
+			phase: state.phase
+		});
+		const shouldExtract =
+			extractionAtSeconds !== null &&
+			!extractionApplied &&
+			elapsed <= extractionAtSeconds &&
+			elapsed + deltaSeconds > extractionAtSeconds;
+		const step = stepMothershipEnergyLift(
+			state,
+			{
+				inflowEnergy: shouldExtract ? extractionEnergy : 0,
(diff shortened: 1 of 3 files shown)


xtreemze:
Integration invariant for the later adapter: this ledger may report `unmetExternalSpend`, but no discrete physical action should be committed before affordability is known. Raid/body launch must pre-authorize the required spend before spawn. Continuous navigation should use available funded propulsion authority rather than apply full engine acceleration and account for an unpaid remainder afterward. External silo attraction is not propulsion and continues regardless of funding. Extraction inflow is credited only after physical delivery. Keep this distinction when #101/#100 are wired into this reserve authority.

xtreemze (commented):
Static review before this becomes next-campaign material:

1. **Authorization needs one explicit transaction rule for multiple demands.** `authorizeDiscreteMothershipSpend()` / `fundContinuousMothershipSpend()` are individually correct against one state snapshot, but two independent callers can both approve/fund against the same reserve and then commit physical effects before the single `stepMothershipEnergyLift()` mutation. The step will cap `externalSpendApplied`, so the ledger stays non-negative while already-committed actions can collectively consume more authority than was actually paid. Before adapter work, define/implement one of: sequential authorization against a caller-owned working reserve, an aggregate reservation transaction, or helpers that accept already-committed spend/protected reserve. Add a deterministic fixture with simultaneous launch + propulsion demand proving total committed physical authority never exceeds funded reserve.

2. **The default Storybook test can false-pass if the lifecycle disappears.** `FiniteRaidHorizon.play` only compares fall<hulk, extraction delay, and 30/60 timing when the corresponding data values are >= 0. If a regression makes `fallAt`/`hulkAt` null, the DOM still exists with `-1` and the behavioral assertions are skipped. For the canonical default fixture, assert dry fall >= 0, dry hulk >= 0, dry fall < dry hulk, replenished fall > dry fall (or explicitly assert the intended no-fall case if calibration changes), and both 30/60 hulk events exist before comparing their timing.

The scalar reserve bounds, finite-input sanitation, phase-threshold ordering, engine-vs-external-attraction ownership, and first-impact/bounce→hulk shape otherwise look coherent for a draft. Keep #102 outside frozen #105; these two points should be resolved before it is folded into a later certification batch.

xtreemze (commented):
Additional authorization-boundary finding, separate from the already-recorded multi-demand transaction issue: malformed discrete spend currently fails open. `positive(NaN | ±Infinity | negative)` becomes `0`, so `authorizeDiscreteMothershipSpend()` returns `authorized: true` for an invalid request (including on an empty non-hulk reserve), and the caller may then commit a physical launch as a free action. For an authority gate, finite/non-negative diagnostic sanitation should not convert invalid action cost into permission. Prefer explicit request validity and `authorized: false` for non-finite/negative discrete costs, with a distinct invalid-input reason/status if useful. Likewise consider failing safe for a non-finite `protectedReserve` rather than silently converting it to zero. Add fixtures for negative/NaN/±Infinity requested spend and malformed protected reserve before adapters use this helper. Deliberate zero-cost actions, if ever needed, should be represented explicitly rather than arising from sanitized bad data.

Issue #107, opened 2026-09-04
Model deterrence taper and prove the anti-farming equilibrium


xtreemze:
Parent: #103
Related: #73, #74, #76, #79, #86, #92

## Purpose

Build an engine-free experimental model for the campaign hinge:

`successful defense → lower attacker expected value → fewer/cheaper commitments → deterrent quiet → defender starvation`

and explicitly test the anti-exploit condition:

`EV(competent defense) > EV(intentional breach farming)`

while preserving the short/medium-term condition:

`EV(active combat) > EV(passive isolation)`

This issue owns modeling and evidence. It does **not** promote final production constants.

## Why this needs a dedicated model

The campaign inversion depends on a narrow but important economic regime. If attacks are too generous, deliberate under-defense becomes rational farming. If attacks are too punitive or deterrence reacts too quickly, combat-derived replenishment stops mattering and the systemic twist feels scripted. If probes themselves can fully sustain a fortress, deterrent quiet never becomes strategically meaningful.

The relationship should be demonstrated over long horizons before live-wave behavior is changed.

## Model state

Start with a small explicit state vector rather than a full game simulation:

### Defender
- stored energy;
- active defense replacement value;
- tower aging/decay burden;
- recent damage/recovery;
- recent breach loss;
- optional stranded/lost recovered energy from #76/#86.

### Attacker population / source system
- perceived target energy/signature;
- recent extraction return;
- recent committed losses;
- breach probability estimate;
- observed time-to-silo / remaining viability;
- travel/operating cost;
- confidence / willingness to commit expensive bodies.

### Raid event
- tier/composition committed;
- launch/embodied energy;
- approach duration;
- viability on arrival;
- extracted energy;
- recoverable value captured by defender;
- attacker net return;
- defender net change.

Keep all formulas injectable/configurable so experiments compare hypotheses rather than bury assumptions in code.

## Deterrence state model

Model at least five qualitative states:

1. `CONTESTED`
2. `ADAPTING`
3. `PROBING`
4. `DETERRENT_QUIET`
5. `STRATEGIC_STARVATION`

Do not make these campaign-index states. They should emerge from rolling expected-value evidence with hysteresis so a single lucky/unlucky raid does not flip the world abruptly.

### Required behavior

- repeated profitable attacks should sustain or increase commitment;
- repeated expensive failures should reduce expensive commitments before reducing all activity;
- R1/scout probes should persist longer than expensive bodies;
- repeated failed probes should eventually produce quiet;
[...]

Issue #103, opened 2026-09-03
Refine causal legibility, conserved conflict economy, and deterrence pacing


xtreemze:
Parent: #29
Related: #72, #73, #74, #76, #79, #86

## Purpose

Refine the existing design without expanding its conceptual surface. The current game already has unusually strong systemic/narrative coherence; the primary design risk is now that players may fail to **predict, perceive, and correctly attribute** the consequences generated by the physics/economy.

This issue therefore owns a focused refinement pass around six questions:

1. Can players predict qualitative physical outcomes before they occur?
2. Can players explain why an important outcome occurred afterward?
3. Can the energy economy remain physically conserved enough that combat does not feel like arbitrary resource creation?
4. Can defender success remain locally desirable without intentional under-defense becoming the dominant long-term strategy?
5. Can deterrence taper into strategic silence without feeling like the game has stopped producing content?
6. Can the defender → raider inversion be understood through reused systems rather than explanatory text?

No new currency, damage type, tech tree, narrative exposition layer, or conventional strategy HUD should be introduced to solve these problems.

---

# 1. Canonical refinement: a causal-legibility contract

Every strategically important mechanic should expose three layers through the world wherever practical:

### Before the event — anticipation
The player should have a qualitative basis for predicting what is likely to happen.

Examples:
- a T3 visibly/soundly commits to a slower, heavier shot before launch;
- a heavy Raider 3 visibly carries inertia and cannot pivot like R1;
- an aging tower visibly approaches degradation before it steps down;
- a narrow opening visibly admits R1 while excluding R3;
- a mothership/fortress visibly dims before loss of operational capability.

### During the event — causation
The important causal chain should remain perceivable while it unfolds.

Examples:
- projectile impulse visibly changes a raider's vector rather than only subtracting HP;
- delay visibly consumes raider viability;
- loose teal energy physically follows available drainage rather than teleporting through walls;
- extraction reverses the established energy-flow direction;
- a mothership that spends energy on movement/deployment visibly pays from the same reserve.

### After the event — residue
Consequences should leave enough temporary or persistent evidence that the player can infer what happened.

Examples:
- pooled/stranded energy shows that fortress geometry obstructed drainage;
- debris/scarring/hulks indicate failed expensive commitments;
- a depleted signature visibly reflects a recently extracted silo;
- a route opened by a destroyed/expired barrier remains spatially obvious;
- a fallen mothership remains as a physical result rather than disappearing as a wave-complete state.

[...]

Issue #72, opened 2026-09-03
Define distinct tactical roles for all three tower and enemy tiers


xtreemze:
Parent: #29
Related: #30, #32, #38, #39, #45, #46, #50

## Purpose

Turn the existing three tower tiers and three enemy tiers into an explicit six-role battlefield system. Each type should justify its existence through a distinct strategic/tactical purpose grounded in Defend's physics-driven resource loop, not simply through larger numbers.

### Tower 1 — barrier / deflector

- only tower that can be directly placed;
- 10×10 footprint, height 3;
- static Cannon body (`mass: 0`, restitution 0);
- no turret, no projectile, no direct damage;
- costs 3000 energy;
- ages after ~25 s and disposes after the ~4 s warning interval;
- meaningful current purpose: block, redirect, delay, funnel, and create edge-ejection opportunities while enemy HP continues to decay.

This is already a distinct archetype and should remain non-gun utility rather than becoming a weak version of Tower 2.

### Tower 2 — rapid interceptor

- obtained by upgrading Tower 1; cumulative spend from a fresh cell is currently 3000 + 6000 = 9000;
- same 10×10 base, plus a ~1.33-wide × 6-high pillar and ~6-wide turret geometry;
- static structure;
- range 90;
- shot interval 208 ms (~4.81 shots/s);
- projectile: approximately 0.8 wide × 0.73 high × 6 long, mass 120, 1760 damage, impulse magnitude 26560;
- nominal direct-damage throughput ≈8462 HP/s and launch-impulse throughput ≈127692 impulse units/s;
- degrades to Tower 1 after its own lifetime.

Natural role: frequent, granular fire for light/medium targets, continuous steering/knockback, low-overkill interception, and coverage of congested approaches.

### Tower 3 — long-range siege / impulse cannon

- obtained by upgrading Tower 2; cumulative spend from a fresh cell is currently 3000 + 6000 + 9000 = 18000;
- same base, plus ~2-wide × 9-high pillar and ~9-wide turret geometry;
- range 135;
- shot interval 702 ms (~1.42 shots/s);
- projectile: approximately 1.36 wide × 1.21 high × 22 long, mass 270, 5940 damage, impulse magnitude 89640;
- nominal direct-damage throughput and launch-impulse throughput are essentially identical to Tower 2 because damage/impulse and cadence all scale by level³;
- individual shots are 3.375× stronger and much heavier, while cadence is 3.375× slower;
- degrades to Tower 2, then eventually Tower 1.

[...]

PR #111, opened 2026-09-04
experiment: model deterrence taper and anti-farming economy

xtreemze:
Refs #107, #103, #73, #74, #76, #92

## Purpose

Make the defender → deterrence → starvation hinge executable as a pure deterministic experiment before changing live wave scheduling or production balance.

## Scope

Exactly two new files on current `master` (`206dc028500f10614504ec0a07b381e5320204ca`):

- `src/js/gameplay/deterrenceEconomy.ts`
- `experiments/storybook/src/Foundations/Strategy/DeterrenceEconomy.stories.ts`

No existing production call site, package/dependency metadata, wave table, balance constant, Babylon/Cannon object, or certification head is modified.

## Model

The engine-free model exposes injectable:

- defender energy / upkeep;
- attacker confidence;
- conserved/capped defender recovery;
- breach loss;
- travel cost;
- R1/R2/R3 commitment/extraction hypotheses;
- defense strength and deliberate leak bias;
- confidence smoothing and hysteresis;
- `CONTESTED → ADAPTING → PROBING → DETERRENT_QUIET → STRATEGIC_STARVATION` states.

Every raid records committed energy, extraction, recovered energy, breach loss, viability and attacker net return. Defender recovery is explicitly bounded by the raid's committed energy so this experiment cannot manufacture unbounded economic value from repeated damage.

All defaults are labeled normalized lab hypotheses. They are not a production-balance proposal.

## Strategy comparison

The Storybook surface compares five policies across deterministic seed sweeps:

1. competent defense;
2. intentional leak/farm;
3. over-fortification;
4. resource-conservative defense;
5. passive isolation.

It visualizes representative reserve trajectories, ruin rate, survival horizon, raids/breaches, attacker ROI and occupancy of the five deterrence states.

The default parameter regime is intentionally only a feasibility witness: competent defense has lower ruin and longer survival than intentional leakage, while aggressive over-fortification is also costly. Controls are provided specifically so reviewers can find failure regimes rather than treating the defaults as solved balance.

## Automated Storybook invariants

The story checks:

- all five policies render;
- default anti-farming comparison passes;
- recovered defender energy never exceeds the corresponding raid commitment;
- defender reserve stays within `[0, max]`.

## Frozen certification boundary

Do **not** add this draft to the current #92 frozen certification cut. It is a later engine-free experimental lane. The frozen exact SHAs for #95/#97/#105 remain unchanged.

## Follow-up evidence

Before this leaves draft status:

- Storybook typecheck/build/test on an appropriate later campaign;
- parameter sweeps beyond the default witness;
- identify regimes with probe farming, runaway energy, over-fast starvation or noisy state oscillation;
- calibrate physical outcome inputs using #72/#86 fixture evidence rather than inventing hidden live-game bonuses;
- decide whether the model remains an experiment, becomes a reusable campaign contract, or is replaced by a better hypothesis.

No production integration should be inferred from this PR alone.

### DIFF
--- a/experiments/storybook/src/Foundations/Strategy/DeterrenceEconomy.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/DeterrenceEconomy.stories.ts
@@ -1,0 +1,311 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	compareDeterrencePolicies,
+	DEFAULT_DETERRENCE_CONFIG,
+	DEFAULT_DETERRENCE_POLICIES,
+	DEFAULT_DETERRENCE_RAIDERS,
+	type DeterrenceModelConfig,
+	type DeterrencePolicyId,
+	type DeterrenceScenarioResult,
+	type DeterrenceStateCounts
+} from "@defend/gameplay/deterrenceEconomy";
+import { createLabShell } from "../../labTheme";
+
+type DeterrenceArgs = {
+	horizon: number;
+	seed: number;
+	seedCount: number;
+	collectionEfficiency: number;
+	confidenceSmoothing: number;
+	breachLossMultiplier: number;
+	starvationThreshold: number;
+};
+
+type PolicyAggregate = {
+	id: DeterrencePolicyId;
+	label: string;
+	meanEndingEnergy: number;
+	meanSurvival: number;
+	ruinRate: number;
+	meanRaids: number;
+	meanBreaches: number;
+	meanAttackerRoi: number;
+	stateCounts: DeterrenceStateCounts;
+	representative: DeterrenceScenarioResult;
+};
+
+function fixed(value: number, digits = 1): string {
+	if (value !== value || value === Infinity || value === -Infinity) return "0";
+	return value.toFixed(digits);
+}
+
+function percent(value: number): string {
+	return `${fixed(value * 100, 0)}%`;
+}
+
+function copyCounts(): DeterrenceStateCounts {
+	return { contested: 0, adapting: 0, probing: 0, quiet: 0, starvation: 0 };
+}
+
+function addCounts(
+	target: DeterrenceStateCounts,
+	source: DeterrenceStateCounts
+): void {
+	target.contested += source.contested;
+	target.adapting += source.adapting;
+	target.probing += source.probing;
+	target.quiet += source.quiet;
+	target.starvation += source.starvation;
+}
+
+function aggregatePolicies(args: DeterrenceArgs): PolicyAggregate[] {
+	const config: DeterrenceModelConfig = {
+		...DEFAULT_DETERRENCE_CONFIG,
+		horizon: args.horizon,
+		collectionEfficiency: args.collectionEfficiency,
+		confidenceSmoothing: args.confidenceSmoothing,
+		breachLossMultiplier: args.breachLossMultiplier,
+		starvationThreshold: args.starvationThreshold
+	};
+	const seedCount = Math.max(1, Math.floor(args.seedCount));
+	const accumulators = DEFAULT_DETERRENCE_POLICIES.map(policy => ({
+		id: policy.id,
+		label: policy.label,
+		endingEnergy: 0,
+		survival: 0,
+		ruins: 0,
+		raids: 0,
+		breaches: 0,
+		attackerRoi: 0,
+		stateCounts: copyCounts(),
+		representative: compareDeterrencePolicies(
+			[policy],
+			config,
+			DEFAULT_DETERRENCE_RAIDERS,
+			args.seed
+		)[0]
+	}));
+
+	for (let offset = 0; offset < seedCount; offset += 1) {
+		const results = compareDeterrencePolicies(
+			DEFAULT_DETERRENCE_POLICIES,
+			config,
+			DEFAULT_DETERRENCE_RAIDERS,
+			args.seed + offset
+		);
+		for (let index = 0; index < results.length; index += 1) {
+			const result = results[index];
+			const acc = accumulators[index];
+			acc.endingEnergy += result.summary.endingDefenderEnergy;
+			acc.survival += result.summary.survivalOpportunities;
+			if (result.summary.ruined) acc.ruins += 1;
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Methodology review before this becomes evidence for #107:

The two-file scope is appropriate and the committed-energy cap is a useful first conservation invariant. Keep this as a feasibility experiment, but I would not treat the current default anti-farming pass as comparative evidence yet.

1. **Synchronize stochastic inputs across policy alternatives.** `compareDeterrencePolicies()` gives every policy the same seed, but `runDeterrenceScenario()` consumes a variable number of RNG draws depending on whether a raid occurs and whether it breaches. Once policies branch, later `random()` values no longer represent the same external event. Common-random-number experimental design is strongest when the *same draws are used for the same stochastic purposes* across alternatives; merely sharing a seed is insufficient when control flow desynchronizes the stream. Prefer a fixed per-opportunity tuple such as `{ raidRoll, tierRoll, breachRoll, viabilityRoll }`, generated/advanced regardless of branch, or separate synchronized streams per stochastic purpose. This makes paired policy differences much more interpretable.

2. **Make silence semantics explicit.** A no-raid opportunity currently always subtracts `idleConfidenceDecay`, while lower confidence itself lowers future raid probability. This creates a self-reinforcing quieting loop without a new observed raid payoff. That may be a useful hypothesis, but it should be modeled as an explicit staleness/forgetting prior rather than implicit negative evidence. A stronger design would separate expected-return estimate from uncertainty/exploration pressure (or at least mean-revert stale confidence toward a configurable target-opportunity prior). Quiet should increase uncertainty enough to motivate cheap probes; probe outcomes should then update expected return. This better matches #107's requirements that probes persist under uncertainty and a visibly rich target can recover interest.

3. **Add the taper-quality outputs already required by #107 before calling the model a proof:** average committed value/raid, cheap-probe fraction, composition entropy/variety, inter-commitment interval, abrupt state-transition count, time strong-defense→quiet, quiet→starvation, and recovery behavior after a surprising successful probe. Also report paired policy differences across synchronized seeds, not only separate means/ruin rates.

4. **Strengthen automated evidence.** The current Storybook assertion proves only that one default parameter witness satisfies `competent > intentional-leak`. Add sweeps/adversarial fixtures that intentionally search for probe farming, runaway energy, immediate starvation, permanent high-intensity raids, noisy state oscillation, and universal policy dominance. The default witness should remain a smoke test, not the anti-farming proof.

5. Add a specific regression fixture for a long interval with no raid outcome. It should establish the chosen semantics explicitly: either expected return remains unchanged while uncertainty grows, or a configured staleness prior moves it in a documented direction. Do not let absence of an attack silently mean observed attacker failure.

Research basis: simulation literature on common random numbers emphasizes synchronized random values for the same stochastic events/purposes across alternative runs; separate streams per stochastic input are a common implementation. Standard recency-weighted value estimation updates from observed rewards, so any no-observation drift here should be an explicit model of staleness/prior dynamics rather than an accidental reward update.

No production integration or #92 frozen-campaign change is warranted yet.

xtreemze:
Conservation review found a structural issue in the current scenario settlement that should be corrected before this model is treated as credible anti-farming/deterrence evidence.

Current #111 computes `extractedEnergy` from theoretical extraction potential × viability even when the defender silo contains less energy, then clamps defender reserve afterward. It also summarizes nominal recovery/breach quantities that may have been clipped by the reserve cap. This can make a poor target economically more valuable than its stored energy permits and weakens the required target-richness → raid-EV relationship from #107.

I opened stacked child #114 with a reusable conserved exchange contract + Storybook witness. Its semantics are:

- actual extraction <= energy physically stored in the target;
- attacker ROI/confidence uses actual extraction, not potential;
- any extra breach severity is dissipative loss rather than extra attacker income;
- recovery is bounded first by attacker embodied energy, then by defender storage headroom;
- unmet extraction, source-limited recovery, and uncollected recovery remain explicit rather than disappearing inside `clamp()`.

Recommended next step for this branch: adopt #114's settlement (or equivalent semantics) in `runDeterrenceScenario`, report actual applied transfers in summaries, then rerun the five-policy seed/parameter sweeps. Keep #111 draft until that conservation boundary is present.

xtreemze (commented):
Attacker-side conservation follow-up after inspecting the runner: `attackerNetReturn = extractedEnergy - raider.committedEnergy - travelCost` treats the entire embodied launch commitment as permanently lost even when a breaching raider retains `remainingViability`. That is inconsistent with the mothership lifecycle in #79, where surviving raiders evacuate and may reintegrate into the mothership reserve. If `committedEnergy` is embodied energy, ROI should account for returned survivor energy (or, equivalently, subtract only permanently lost embodied energy): `net = actual extraction + returned surviving embodied - committed embodied - travel/operating cost`. The returned fraction should be tied to physical survivor/recovery state, not a hidden tier bonus. This should be resolved together with #114's capture-eligible recovery partition; otherwise the model can simultaneously undervalue successful durable raids and double-account defender recovery from surviving bodies. I expanded the conservation note on #107 with the required partition/invariant.

xtreemze:
Static model review found two structural problems that need correction before this experiment is useful as evidence:

1. `extractedEnergy` is currently computed from raider viability without capping it to the defender reserve actually available after upkeep. The defender reserve later clamps at zero, so the attacker can be credited with energy that did not exist. That violates the conserved-energy premise.
2. Raid frequency/tier selection depends on deterrence state but not on target richness. A nearly empty fortress can therefore still receive expensive commitments if confidence remains high, contrary to #107/#76's requirement that poor targets should not attract irrationally expensive raids.

I am treating these as model defects, not balance-tuning questions. The draft should be amended so extraction is physically capped, breach dissipation cannot make extraction exceed defender loss, recovery is tied to damaged/removed embodied raider value, and target energy/signature influences both willingness to raid and high-tier commitment. Add explicit invariants for these conservation relationships.

xtreemze:
The deterrence design has now been decomposed enough that this draft should remain a historical/feasibility parent rather than absorb further assumptions directly. #128 defines the candidate successor architecture: #114 authoritative conserved settlement, #125 capital-at-risk comparison, #127 contextual attacker beliefs, then a thin planner/long-horizon runner. Do not promote this PR's current tier selection/confidence model into production while those narrower seams remain unresolved.

Issue #109, opened 2026-09-04
Fix exact-cost placement/upgrade affordability after gameplay baseline certification


xtreemze:
Parent: #108
Related: #30, #32, #43, #95

## Observed behavior

The current placement path requires:

`currentBalance > baseCost`

and treats:

`currentBalance <= baseCost`

as insufficient funds. The upgrade path uses the same strict-greater-than pattern against `baseCost * (currentLevel + 1)`.

This means a player holding **exactly** the displayed/required cost cannot perform the action.

That is counter to ordinary affordability semantics and weakens direct-manipulation intuitiveness: the visible resource state says the action is affordable while the interaction rejects it.

## Why this is separate

#95 intentionally preserves current gameplay for exact-head baseline certification, including current tower placement/upgrade behavior. Do not alter #95 merely to fix this.

First characterize/certify the legacy behavior under #30/#92/#95. Then fix the semantic boundary in a narrow follow-up PR.

## Intended rule

Unless baseline evidence reveals a deliberate minimum-reserve mechanic that is not currently documented, affordability should be:

`currentBalance >= cost`

with rejection only when:

`currentBalance < cost`.

Apply the same rule consistently to placement and upgrades.

## Required implementation properties

- centralize affordability in a pure/testable helper rather than duplicating comparison operators in pointer handlers;
- preserve the actual charge amount and all unrelated tower behavior;
- no hidden one-unit reserve;
- no change to tower tier costs;
- no change to level-3 semantics in this PR;
- failed actions continue to emit the appropriate feedback path from #108;
- exact-cost success must leave the reserve at zero without bypassing the normal defeat/update logic.

## Certification cases

For placement and each supported upgrade transition, test:

1. balance = cost - 1 → rejected, no charge/object mutation;
2. balance = cost → succeeds exactly once, resulting balance 0;
3. balance = cost + 1 → succeeds exactly once, resulting balance 1;
4. repeated input does not double-charge;
5. occupied/invalid targets remain rejected independently of affordability;
6. non-tower taps remain harmless.

Record the legacy exact-cost behavior first on #95, then certify the fix against the follow-up head.

## Success criterion

The game never reports or visually implies an action cost that the player can exactly meet while silently requiring an undisclosed additional unit of reserve.

xtreemze:
Draft #112 now exposes the affordability boundary as a pure caller-owned policy without changing production behavior.

`TowerAffordabilityRule` has:
- `legacy-strict` → current `balance > cost` behavior;
[...]

Issue #108, opened 2026-09-04
Operationalize ludic/intuitiveness findings with direct-manipulation and first-contact comprehension gates


xtreemze:
Parent: #103
Related: #29, #30, #33, #56, #76, #77, #86, #92, #95

## Diagnosis

Defend's core play is **operationally intuitive but systemically opaque**.

The direct interaction is already strong: touch/clicking the world is the primary interface, construction/upgrade are spatial acts, and the game avoids menu-heavy tower-defense control. The ludic structure is also unusually coherent: one energy reserve binds health, construction and combat recovery; physics permits destruction, delay, obstruction, redirection and ejection; temporary defenses prevent a permanently solved board.

The weakness is not lack of depth. It is that several defining causal rules are difficult for a first-time player to infer from the world quickly enough to act intentionally.

The implementation/design work under #103 should therefore be judged by a stricter criterion than correctness:

> Can a player predict an action's qualitative consequence before committing, and explain the result afterward, without requiring a conventional strategy HUD or tutorial recipe?

## Scope

This issue owns the **player-facing operationalization** of #103 for the defensive play loop:

1. placement/upgrade feedforward;
2. explicit but world-grounded rejection feedback;
3. first-contact encounter sequencing that teaches transferable causal relationships;
4. comprehension/interaction metrics;
5. intervention-cadence checks so temporary defenses create decisions rather than repetitive maintenance.

It does **not** own final hex topology (#56), final physical energy rendering (#76/#77), final balance, mothership controls, or broad input-platform migration (#33).

---

## 1. Direct-manipulation contract

Every build/upgrade attempt should have three stages.

### Before commitment: feedforward

When the pointer/touch target resolves to a candidate build cell or tower, expose enough state to answer:

- where will the action occur?
- what object/tier will result?
- is the target valid?
- is it occupied/blocked?
- is the action affordable?
- for an existing tower, is this an upgrade, maintenance/refresh action, or no-op?

Prefer world feedback: projected footprint, ground response, tower-state response, subtle range/physical influence preview where useful. Do not introduce a permanent build palette merely to explain these states.

### During commitment: causal response

The action should visibly connect resource spend to the constructed/reconstructed object. When a build is rejected, the reason should be distinguishable rather than becoming a silent tap.

### After commitment: residue

The player should be able to identify what changed and, for temporary structures, whether the action altered level, lifetime/service state, or neither.

---

## 2. No silent rejection

Audit every direct-world interaction path for rejected input.

[...]

PR #112, opened 2026-09-04
Add explicit tower interaction intent and rejection states

xtreemze:
Refs #108
Related: #109, #95, #92

## Purpose

Introduce the pure interaction-state seam requested by #108 before changing Babylon pointer handlers or player-facing presentation.

Current production interaction has several semantically different outcomes that are either silent or coupled directly to scene/audio/color mutation. This PR makes the intended action and rejection cause explicit without changing live gameplay.

## `src/js/gameplay/towerInteraction.ts`

Adds an engine-free classifier for a pointer/touch candidate.

It distinguishes:

- placement;
- upgrade;
- max-tier rebuild;
- future maintenance;
- no-op;
- camera-owned gesture;
- occupied cell;
- unaffordable action;
- protected/core space;
- invalid terrain;
- outside arena;
- stale/disposed target;
- invalid tower level;
- max-level rejection.

The result contains semantic intent, allowed/rejected/ignored disposition, explicit reason, requested cost, balance before/projected balance after, and source/resulting tower level.

The module performs no Babylon picking, scene mutation, audio, color feedback, tower construction/disposal, maintenance, or authoritative reserve mutation.

## Deliberately unresolved legacy boundaries

This PR does **not** silently fix behavior that #95/#109 still need to characterize.

### Exact-cost affordability

Current production placement/upgrade requires `balance > cost`, so a balance exactly equal to the displayed cost is rejected.

The classifier therefore makes affordability caller-owned:

- `legacy-strict` → `balance > cost`;
- `inclusive` → `balance >= cost`.

The default Storybook fixture uses `legacy-strict`. #109 can switch a later production adapter to `inclusive` only after the #95 baseline is recorded.

### Level-3 tap semantics

Current production reconstructs level 3 when an affordable level-3 tower is tapped. Whether that should remain a rebuild, become maintenance/renovation, or become a no-op is a separate design decision.

The classifier therefore exposes `legacy-rebuild`, `maintenance`, and `no-op`. No canonical change is made here.

## Storybook playground

Adds `Arena/Interaction/Tower Intent` with a deterministic eleven-state matrix:

1. open ground;
2. exact-cost placement;
3. below-cost placement;
4. occupied cell;
5. protected core;
6. level-1 tower upgrade;
7. level-3 tower policy;
8. invalid terrain;
9. outside arena;
10. camera-owned gesture;
11. stale/disposed target.

Controls expose reserve, affordability policy and max-tier policy. Cards show classified intent, reason, cost, projected reserve and level transition.

The interaction test verifies the exact-cost policy boundary, true under-cost rejection under both policies, occupied/protected attribution, outside-arena attribution, lower-tier upgrade intent, camera arbitration, stale-target classification and all three max-tier policies.

## Scope

Base: current `master` `206dc028500f10614504ec0a07b381e5320204ca`.

Compared with `master`:

- exactly 2 added files;
- no live pointer/input call-site changes;
- no tower/economy/balance mutation;
- no changes to #95/#97/#105 frozen certification heads.

Current head after Storybook coverage hardening: `bb78ef4132b467c21983fd43ecd3cd619e829f8a`.

## Local certification

Keep draft until a later post-freeze campaign can run.

### Root compatibility

- TypeScript 3.2 accepts `towerInteraction.ts`;
- NaN/Infinity/negative numeric inputs remain finite/non-negative in returned previews;
- `legacy-strict` preserves exact-cost rejection;
- `inclusive` permits exact-cost and projects reserve to zero;
- a true under-cost request remains rejected under both rules;
- occupied/protected/invalid/outside/stale reasons remain independent of affordability;
- lower-tier tower resolves to one-step upgrade;
- max-tier behavior follows the selected explicit policy;
- camera-owned input never becomes a world rejection or charge.

### Storybook

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Exercise `Arena/Interaction/Tower Intent` across low/exact/high reserve and all max-tier modes.

## Follow-up

After #95 records legacy interaction behavior:

1. #109 may use this seam/helper to implement ordinary exact-cost affordability without duplicating comparison operators;
2. a Babylon adapter can classify protected/occupied/terrain/outside/stale/camera states before mutation and map each reason to world-grounded multimodal feedback;
3. max-tier rebuild versus maintenance/no-op should be decided explicitly, with #96/#105 maintenance semantics available as a future integration reference;
4. only then should the first-contact/browser interaction fixture from #108 exercise real pointer/touch arbitration.

This PR is post-freeze work and must not be inserted into the current #92 campaign.

### DIFF
--- a/experiments/storybook/src/Arena/Interaction/TowerIntent.stories.ts
+++ b/experiments/storybook/src/Arena/Interaction/TowerIntent.stories.ts
@@ -58,6 +58,12 @@ function scenarios(args: TowerIntentArgs): Scenario[] {
 			request: request(args, { balance: BASE_COST })
 		},
 		{
+			id: "under-cost",
+			label: "Below cost",
+			description: "A true insufficient-reserve case remains rejected under both affordability policies.",
+			request: request(args, { balance: BASE_COST - 1 })
+		},
+		{
 			id: "occupied",
 			label: "Occupied cell",
 			description: "Affordability must not hide an independent spatial rejection.",
@@ -96,6 +102,12 @@ function scenarios(args: TowerIntentArgs): Scenario[] {
 			request: request(args, { targetKind: "invalid-terrain" })
 		},
 		{
+			id: "outside-arena",
+			label: "Outside arena",
+			description: "A world hit outside the playable surface is distinguishable from invalid terrain.",
+			request: request(args, { targetKind: "outside-arena" })
+		},
+		{
 			id: "camera-gesture",
 			label: "Camera owns gesture",
 			description: "Input arbitration is ignored by world placement rather than reported as an economic failure.",
@@ -111,7 +123,7 @@ function scenarios(args: TowerIntentArgs): Scenario[] {
 }
 
 function statusLabel(result: TowerInteractionPreview): string {
-	if (result.disposition === "allowed") return result.intent.toUpper();
+	if (result.disposition === "allowed") return result.intent.toUpperCase();
 	if (result.disposition === "ignored") return "CAMERA";
 	return result.reason.replace(/-/g, " ").toUpperCase();
 }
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Static review found one fail-open edge that should be corrected before this becomes the semantic gate for live mutation. `requestedCost` is sanitized with `nonnegative()`, so a negative, NaN, or infinite cost becomes `0`. `permitted()` can then approve a free placement/upgrade/rebuild/maintenance (under `inclusive`, even at zero balance; under `legacy-strict`, whenever balance > 0). The PR explicitly intends to classify before mutation, so malformed economic input should fail closed rather than normalize into a valid free action. Prefer an explicit `invalid-cost` rejection (or equivalent validity check before affordability), and add fixtures for negative/NaN/±Infinity cost. Sanitizing preview output to finite values is still useful, but it should not convert invalid authority into permission.

xtreemze:
Canonical max-tier behavior is now tracked explicitly in #130. Keep #112's `legacy-rebuild` / `maintenance` / `no-op` policy split intact; do not choose a production behavior inside this pure classifier. #95 should characterize the legacy reconstruction first, and #96/#105 maintenance semantics should be certified before any production adapter selects maintenance.

Issue #86, opened 2026-09-03
Make defensive geometry shape energy drainage and raider locomotion


xtreemze:
Parent: #29
Related: #72, #76, #82, #83

## Purpose

Make surface geometry authoritative for both recovered teal-energy flow and raider movement.

A fortress should not be allowed to become economically magical: walls and tower foundations that block bodies must also block loose surface energy. The defender therefore needs to preserve drainage corridors to the silo while still denying useful raider routes.

## Canonical spatial rule

- walls/towers obstruct free teal-energy flow;
- energy may route around geometry only where a physical opening exists;
- a sealed barrier can strand/pool recovered energy outside the silo;
- destroying, expiring, moving, or melting geometry can release previously blocked energy;
- terrain slope affects both fluid and spherical raider bodies;
- the silo attraction field biases flow but does not grant obstacle immunity.

This creates the intended trade-off:

**harder to penetrate vs easier to drain.**

### R1 — Navigator / swarm
- highest local path-awareness;
- prefers openings, low terrain and low-collision-cost routes;
- exploits narrow passages enabled by its small diameter;
- avoids needless wall impacts when a nearby route exists.

### R2 — Breaker
- coarse route planning;
- trades detour cost against deliberate impact through temporary/breakable geometry;
- uses mass and repeated collision to compromise gates and walls;
- follows breaches once created.

### R3 — Titan / siege boulder
- lowest detailed pathfinding;
- strongest dependence on slope, inertia and initial insertion vector;
- prefers broad/downhill/direct approaches;
- accepts major collisions instead of elegantly solving mazes;
- cannot use narrow openings merely because a navigation graph contains them.

## Shared-geometry consequences

- a narrow drainage gate may pass energy and R1 but exclude R3;
- a wider high-throughput channel may become an R2/R3 attack lane;
- multiple small drains can improve economic resilience while fragmenting R1 approach paths;
- T1 walls may successfully redirect a Titan while trapping recovered energy outside the fortress;
- eruptions can simultaneously create a new raid route and a new energy-drain route;
- visible drainage streams can reveal openings to a raider player/AI.

## Implementation architecture

Do not pathfind every rendered droplet.

Energy should use low-count authoritative aggregates plus a coarse potential/height/obstacle representation or bounded local steering. If no physical path exists, aggregate energy pools instead of performing intelligent maze solving.

Raider path planning should update on topology changes/bounded cadence, not every render frame.

## Required fixtures

Energy:
1. open field;
[...]

PR #114, opened 2026-09-04
experiment: enforce conserved raid exchange against target richness

xtreemze:
Refs #107, #103
Parent experiment: #111
Related: #74, #76, #86, #102

## Purpose

Add a narrow conserved-energy settlement seam underneath the deterrence experiment before #111's attacker ROI is treated as meaningful.

Review of #111 found a structural issue: theoretical extraction potential is currently credited to the attacker even when the defender silo contains less energy, while defender recovery/breach totals can describe amounts later clipped by a reserve clamp. That weakens the required `rich target -> more attractive / poor target -> less attractive` loop and can manufacture apparent energy in the headless model.

This child deliberately does **not** rewrite #111's scenario runner yet. It adds the reusable settlement contract and a deterministic witness fixture so the parent can adopt the semantics in one focused follow-up.

## `src/js/gameplay/conservedRaidExchange.ts`

Adds an engine-free exchange contract with explicit event ordering for opportunity-level modeling:

1. breach extraction draws only from energy already stored in the defender silo;
2. optional additional breach harm is dissipative collateral and never becomes attacker income;
3. combat recovery is then collected into remaining defender storage.

The result exposes:

- requested vs actual extraction;
- unmet extraction caused by insufficient target energy;
- requested vs actual dissipative collateral loss;
- requested recovery vs attacker embodied/source-available recovery;
- source shortfall when recovery demand exceeds attacker embodied energy;
- actually collected recovery vs storage-limited uncollected recovery;
- actual defender energy lost;
- before/after reserve states.

`raidAttackerNetReturn()` explicitly uses **actual extracted energy**, not theoretical extraction potential.

`raidTargetRichness()` exposes a lossy reserve/capacity ratio for experiments/signatures. It is not authoritative state and not a difficulty level.

## Why collateral is dissipative

A breach may eventually have effects beyond transferred energy, but a scalar multiplier below 1 must not let an attacker receive more energy than the defender loses. This child therefore represents any extra breach severity as non-negative dissipation after actual extraction. Final production consequences remain an open balance/design question.

## Storybook playground

Adds `Foundations/Strategy/Conserved Raid Exchange`.

The default fixture applies the same requested raid to:

- a rich/full target;
- a poor 5,000-energy target.

The poor target cannot yield the same 18,000 requested extraction, so actual attacker return falls automatically as target richness falls.

Additional deterministic stress cases verify:

- a nearly-full silo reports only recovery that actually fits, with the remainder explicitly uncollected;
- requested recovery cannot exceed attacker embodied energy even when defender storage is empty.

The story asserts conservation identities and creates no RAF, timer, Babylon scene, Worker or AudioContext.

## Integration requirement for #111

Before #111 is considered a credible anti-farming/deterrence model, its raid settlement should consume these semantics (or an equivalent conservation-preserving implementation):

- use actual target-bounded extraction for attacker ROI/confidence;
- use actually applied recovery/loss in summaries rather than pre-clamp nominal values;
- record unmet extraction and uncollected/source-limited recovery;
- make low reserve naturally reduce raid value instead of relying on an abstract starvation-state switch alone.

That directly satisfies #107's requirement that visibly rich targets recover interest more readily and poor targets do not receive irrationally valuable commitments.

## Scope

Relative to #111 head `79a260e45bc70126bdd2ea9975101b09214a55a2`:

- 2 added files;
- no edits to #111's existing scenario runner;
- no final balance constants.

## Validation

Keep draft until a later Storybook/root campaign can run:

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Also test zero capacity/reserve, extraction greater than reserve, recovery greater than embodied attacker energy, recovery greater than storage headroom, and non-finite inputs.

This remains post-freeze work and must not alter the active #92 exact-SHA campaign.

### DIFF
--- a/experiments/storybook/src/Foundations/Strategy/ConservedRaidExchange.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/ConservedRaidExchange.stories.ts
@@ -54,8 +54,10 @@ const meta = {
 		travelCost: { control: { type: "range", min: 0, max: 5000, step: 100 } }
 	},
 	render: (args: ExchangeArgs) => {
+		const richReserve = Math.max(args.richReserve, args.poorReserve);
+		const poorReserve = Math.min(args.richReserve, args.poorReserve);
 		const rich = settleConservedRaidExchange({
-			defenderReserve: args.richReserve,
+			defenderReserve: richReserve,
 			defenderCapacity: args.capacity,
 			requestedExtractionEnergy: args.requestedExtraction,
 			attackerEmbodiedEnergy: args.attackerEmbodiedEnergy,
@@ -63,7 +65,7 @@ const meta = {
 			collateralDissipationRatio: args.collateralDissipationRatio
 		});
 		const poor = settleConservedRaidExchange({
-			defenderReserve: args.poorReserve,
+			defenderReserve: poorReserve,
 			defenderCapacity: args.capacity,
 			requestedExtractionEnergy: args.requestedExtraction,
 			attackerEmbodiedEnergy: args.attackerEmbodiedEnergy,
@@ -138,8 +140,8 @@ const meta = {
 				.exchange-note { margin-top:14px; padding:12px; border:1px solid rgba(73,215,209,.18); font-size:11px; line-height:1.55; opacity:.76; }
 			</style>
 			<div class="exchange-grid">
-				${card("rich target", args.richReserve, rich.extractedEnergy, rich.unmetExtractionEnergy, rich.recoveredEnergy, rich.reserveAfter, richReturn)}
-				${card("poor target", args.poorReserve, poor.extractedEnergy, poor.unmetExtractionEnergy, poor.recoveredEnergy, poor.reserveAfter, poorReturn)}
+				${card("rich target", richReserve, rich.extractedEnergy, rich.unmetExtractionEnergy, rich.recoveredEnergy, rich.reserveAfter, richReturn)}
+				${card("poor target", poorReserve, poor.extractedEnergy, poor.unmetExtractionEnergy, poor.recoveredEnergy, poor.reserveAfter, poorReturn)}
 			</div>
 			<div class="exchange-note"
 				data-rich-extracted="${rich.extractedEnergy}"
--- a/experiments/storybook/src/Foundations/Strategy/ConservedRaidExchange.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/ConservedRaidExchange.stories.ts
@@ -1,0 +1,198 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	raidAttackerNetReturn,
+	raidTargetRichness,
+	settleConservedRaidExchange
+} from "@defend/gameplay/conservedRaidExchange";
+import { createLabShell } from "../../labTheme";
+
+type ExchangeArgs = {
+	capacity: number;
+	richReserve: number;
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
The conservation bounds are a useful correction to #111: actual extraction is target-reserve-bounded, collateral is dissipative rather than attacker income, recovery is bounded by attacker embodied energy and storage headroom, and `raidAttackerNetReturn()` uses actual extraction. Keep those invariants.

One causal issue should be resolved before #111 consumes this helper as its opportunity settlement:

**Do not make `extraction → collateral → recovery` the canonical raid chronology.** In live play, recoverable energy is liberated by combat while raiders approach, but #97 makes reserve credit depend on physical packet arrival at the silo. A surviving raider may breach before some packets arrive, after some arrive, or between multiple arrivals. Those cases produce different available reserve at breach and therefore different actual extraction/ROI.

Example: begin at 5,000 reserve, liberate 4,500 recoverable energy, then request an 18,000 breach. If the 4,500 reaches the silo before breach, up to 9,500 can be extracted; if it arrives afterward, only 5,000 can be extracted and the later collection may refill the defender. The current aggregate helper always chooses the latter chronology. That is a modeling hypothesis, not a conservation identity.

Recommended boundary:

- split the primitive into independently composable transitions such as `collectRaidRecovery(...)` and `settleRaidBreach(...)`, or accept an explicit ordered event list/timestamps;
- let the headless deterrence experiment choose and document an opportunity-level ordering hypothesis;
- later calibrate that ordering/distribution from #97/#86 physical-flow timing rather than baking it into the ledger;
- expose `reserveAtBreach`/pending recovery in the #111 model so target richness means the energy physically available when extraction occurs, not merely reserve at opportunity start.

Also add a chronology witness to Storybook: the same starting reserve/extraction/recovery quantities with recovery-before-breach versus recovery-after-breach should both conserve energy but intentionally produce different extracted energy/end reserve. This will prevent a future refactor from confusing conservation with commutativity.

Secondary checks for the later integration:
- keep `defenderEnergyLost = extraction + dissipative collateral` separate from net reserve delta (`-lost + collected recovery`), as the current result correctly does;
- ensure #111's `committedEnergy` semantics remain clear: if commitment is a sunk launch expenditure, subtracting it once in attacker net return is correct; if surviving bodies return embodied value, a later model will need an explicit returned-energy term rather than changing this exchange silently;
- continue treating `raidTargetRichness()` as a lossy perceptual/decision feature, not authoritative state.

#114 should remain draft/post-freeze. With chronology made explicit, it becomes a strong foundation for the synchronized-RNG/uncertainty revisions already requested on #111.

xtreemze (commented):
The target-side conservation is a useful correction, but the recovery-source invariant needs to be made unambiguous before #111 treats this as authoritative settlement. `requestedRecoveryEnergy` is bounded by `attackerEmbodiedEnergy`; if that field means total raid/committed embodied energy, a mixed outcome can double-account a surviving breacher: it can remain alive to extract defender energy while the defender also recovers energy from the same embodied source. The interface comment says “available to be captured,” which is narrower than the field name and some surrounding wording. Please make capture eligibility explicit—e.g. rename to `recoverableAttackerEnergy`/`captureEligibleAttackerEnergy`, or model surviving/escaped versus destroyed/captured embodied energy—and add a mixed-survivor fixture proving `recovered + surviving/escaped embodied energy <= committed embodied energy`. That will make the conservation claim closed under the actual raid outcomes #107 needs, rather than relying on callers to infer an unstated partition.

Issue #33, opened 2026-09-03
Refresh rendering, input, audio, PWA, and accessibility without weakening direct world interaction


xtreemze:
Parent: #29
Related: #30, #31

## Goal

Bring Defend's presentation and interaction layer up to a modern standard while preserving its procedural, physics-first, low-interface identity.

## Design constraints

- The battlefield remains the primary interface.
- Pointer/touch input must continue to support direct tower placement and upgrading.
- Camera controls must not fight placement/upgrade gestures.
- Keyboard/controller accessibility may be added, but should map naturally onto the same spatial actions rather than create a separate game mode.
- Procedural/abstract visuals are preferred over asset-heavy replacement art unless there is a clear gameplay reason.
- Spatial/procedural sound remains valuable; audio effects should reflect game state without becoming required for comprehension.
- Visual quality should degrade before simulation fidelity on constrained hardware.
- The game should remain installable/offline-capable as a modern PWA where practical.

### Rendering
- modern Babylon rendering pipeline and physically sensible defaults;
- resolution/device-pixel-ratio scaling and adaptive quality;
- glow/bloom/AA choices that retain the neon/wireframe identity without excessive fill cost;
- instancing/thin instances/particles appropriate to the current entity counts;
- clearer telegraphing of enemy health, bank energy, tower aging, range, target acquisition, and projectile impulse without UI clutter.

### Input
- Pointer Events and touch behavior on current browsers;
- robust gesture arbitration between orbit/zoom and placement;
- keyboard/controller navigation/placement as an accessibility enhancement;
- reduced-motion and input feedback preferences.

### Audio
- isolate the procedural sound API from gameplay state;
- modern Web Audio/Tone compatibility where necessary;
- preserve spatial listener updates and event-driven synthesis;
- add a master mute/volume path that does not alter simulation.

### PWA / platform
- replace obsolete service-worker/PWA generation flow;
- current manifest/icons/install behavior;
- offline caching strategy for the game shell;
- responsive/mobile safe-area behavior;
- avoid tracking generated platform packages unless there is a strong release reason.

### Accessibility
- visible/focusable start/help/restart controls;
- semantic announcements for wave, energy, victory/defeat where useful;
- color is not the only carrier of critical state;
- scalable text and usable contrast without turning the playfield into a conventional HUD.

## Parallelization

[...]

PR #115, opened 2026-09-04
Add deterministic direct-manipulation outcome and rejection contract

xtreemze:
Refs #108, #103
Related: #30, #33, #95, #109

## Purpose

Operationalize the defensive direct-manipulation findings without changing the legacy placement/upgrade runtime yet.

Current `master` shows an important asymmetry:

- unaffordable placement/upgrade already produces some sound/indicator/currency response;
- occupied/protected placement collapses into no mutation with no semantic rejection result;
- non-ground/non-tower taps are simply ignored;
- max-tier upgrade currently reconstructs T3;
- affordability still uses the legacy strict-`>` rule that #95 must characterize before #109 intentionally changes it.

This PR adds a pure classifier so presentation can eventually explain authoritative outcomes without moving ownership for picking, occupancy, economy, gesture arbitration, or tower mutation.

## `src/js/gameplay/directManipulation.ts`

Adds an engine-free direct-world interaction contract.

Caller-owned facts:

- target kind (`ground / tower / other / none`);
- whether camera or world action won gesture arbitration;
- stale/disposed target state;
- terrain validity;
- occupancy;
- protected/core status;
- **affordability verdict supplied by the economy authority**;
- required energy for feedforward;
- current/max tower level;
- upgrade vs maintenance intent;
- whether maintenance is meaningful.

The classifier returns exactly one:

- accepted placement;
- accepted upgrade;
- accepted maintenance;
- unaffordable;
- occupied;
- protected;
- invalid terrain;
- maximum semantic state;
- no service needed;
- camera-owned gesture;
- stale target;
- unsupported target.

Each outcome also carries:

- semantic action;
- accepted/rejected state;
- cue family;
- current/resulting tower level;
- required energy;
- durable residue (`new-tower / higher-tier / serviced / unchanged`).

### Economy boundary

The module **does not calculate affordability**. It receives `affordable: boolean` from the caller. Therefore it does not encode the current strict `>` comparison and does not pre-empt #109's eventual exact-cost correction.

### Rejection precedence

The contract favors the most causally useful explanation. For example, an already-max-tier tower reports `maximum-state` even if the supplied affordability fact is false; the player should not be told that more energy would make an impossible upgrade valid.

## Storybook playground

Adds `Foundations/Interaction/Direct Manipulation`.

The matrix renders eleven representative cases:

1. valid placement;
2. unaffordable placement;
3. occupied cell;
4. protected/core cell;
5. invalid terrain;
6. camera gesture ownership;
7. stale/disposed target;
8. valid upgrade;
9. maximum tower state;
10. maintenance available;
11. no service needed.

Feedback is deliberately redundant:

- glyph;
- border/pattern/shape treatment;
- explicit wording;
- action/reason/residue facts.

Color and sound are therefore not required to distinguish critical states.

The Storybook assertion pins all major rejection classes, exactly three accepted actions in the default matrix, and max-tier precedence.

No RAF, timer, Babylon scene, Worker, AudioContext, gameplay mutation or global listener is created.

## Scope

Relative to current `master` `206dc028500f10614504ec0a07b381e5320204ca`:

- 2 commits ahead / 0 behind;
- 2 added files;
- no live `pick.ts` / `upgradeTower.ts` changes;
- no affordability-semantic change;
- no camera/input ownership change;
- no balance constants.

## Validation

Post-freeze draft. Later root + Storybook checks:

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Also add focused root fixtures for every rejection reason, priority collisions (protected+occupied, maximum+unaffordable, stale+otherwise-valid), malformed numeric input, and maintenance outcomes.

## Follow-up

After #95 captures current behavior and this contract is certified:

1. adapt placement/upgrade preview/presentation to call the classifier before mutation;
2. keep #109's affordability policy as a separate intentional behavior change;
3. give each rejection cue a world-grounded accessible presentation;
4. record attempts/rejections/camera conflicts for #108's first-build and attribution gates;
5. integrate certified conserved-energy transfer visuals for construction/recovery rather than inventing a UI-only economy explanation.

This PR is post-freeze and must not change #92's active exact-SHA campaign.

### DIFF
--- a/experiments/storybook/src/Foundations/Interaction/DirectManipulation.stories.ts
+++ b/experiments/storybook/src/Foundations/Interaction/DirectManipulation.stories.ts
@@ -1,0 +1,201 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	classifyDirectManipulation,
+	type DirectManipulationCandidate,
+	type DirectManipulationOutcome
+} from "@defend/gameplay/directManipulation";
+import { createLabShell } from "../../labTheme";
+
+type Scenario = {
+	id: string;
+	label: string;
+	candidate: DirectManipulationCandidate;
+};
+
+function baseCandidate(): DirectManipulationCandidate {
+	return {
+		targetKind: "ground",
+		gestureOwner: "world",
+		targetStale: false,
+		terrainValid: true,
+		occupied: false,
+		protected: false,
+		affordable: true,
+		requiredEnergy: 3000,
+		currentTowerLevel: null,
+		maximumTowerLevel: 3,
+		requestedTowerAction: "upgrade",
+		maintenanceAvailable: false
+	};
+}
+
+function scenarios(): Scenario[] {
+	const validPlace = baseCandidate();
+	const unaffordable = { ...baseCandidate(), affordable: false };
+	const occupied = { ...baseCandidate(), occupied: true };
+	const protectedCell = { ...baseCandidate(), protected: true };
+	const invalidTerrain = { ...baseCandidate(), terrainValid: false };
+	const cameraOwned = { ...baseCandidate(), gestureOwner: "camera" as const };
+	const stale = { ...baseCandidate(), targetStale: true };
+	const validUpgrade = {
+		...baseCandidate(),
+		targetKind: "tower" as const,
+		currentTowerLevel: 2,
+		requiredEnergy: 9000
+	};
+	const maximum = {
+		...validUpgrade,
+		currentTowerLevel: 3,
+		affordable: false
+	};
+	const service = {
+		...validUpgrade,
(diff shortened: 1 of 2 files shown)


xtreemze:
Closing as superseded by concurrent #112 + fail-closed child #117.

#112 already owns the same interaction-intent/rejection seam with more complete legacy/future policy representation (explicit exact-cost policy, max-tier rebuild/maintenance/no-op choices, outside-arena and invalid-level states). #117 additionally fixes malformed-cost authorization fail-open behavior.

The one extra concept here, `no-service-needed`, should be reconsidered later when #96/#105 maintenance semantics are integrated into the authoritative interaction classifier; it is not sufficient reason to maintain a duplicate contract.

No production code depends on this branch. Observation work from #116 will be rebuilt against #117 instead.

PR #116, opened 2026-09-04
Add passive direct-manipulation comprehension metrics

xtreemze:
Refs #108, #103
Parent: #115
Related: #92, #95, #109

## Purpose

Add the passive measurement layer required by #108's first-build/rejection-attribution and maintenance-repetition gates without instrumenting or changing the production runtime yet.

This PR is stacked on #115 so it consumes the already-classified direct-manipulation outcome rather than rediscovering placement, camera, occupancy or economy semantics.

## `src/js/gameplay/directManipulationObservation.ts`

Adds an engine-free append-only observation contract.

Each observation supplies:

- session-relative timestamp;
- optional cell identity;
- optional semantic role identity;
- caller-owned coarse tactical-context fingerprint;
- intervention kind (`new-build / replacement / upgrade / maintenance / other`);
- the authoritative #115 classified outcome.

The observer derives only evidence:

- total attempts;
- accepted/rejected actions;
- acceptance rate;
- rejection counts by semantic reason;
- camera/placement conflicts;
- rejected attempts before first successful build;
- time to first accepted action;
- time to first successful build;
- accepted replacement count;
- same-cell/same-role replacements made under unchanged tactical context;
- repetition rate for those replacements.

### Tactical-context boundary

The observer never reads gameplay state. `tacticalContextKey` is supplied by the caller as a deliberately coarse fingerprint describing whether the tactical situation materially changed.

A replacement is counted as rote/input-tax only when:

- it is an accepted `replacement` intervention;
- the same cell and role had a prior accepted action;
- the caller's tactical-context fingerprint is unchanged.

If enemy mix, topology, reserve pressure, breach state, or another meaningful factor changes and the caller changes the fingerprint, the replacement is not counted as rote merely because it happens on the same cell.

### Timing robustness

Session time is monotonic inside the observer. An out-of-order observation cannot move first-success timing backward.

## Storybook playground

Adds `Foundations/Interaction/Direct Manipulation Observation`.

It compares two deterministic illustrative sessions:

### Legible first contact
- one occupied-cell rejection;
- one camera arbitration conflict;
- successful first build at 4.2 s;
- two later replacements, only one under unchanged context;
- later upgrade after threat context changes.

### High-friction / rote maintenance
- four rejected attempts before first build;
- first build later;
- repeated same-cell/same-role replacements under unchanged context.

The fixture does **not** assert those illustrative numbers as final UX targets. It verifies the measurement semantics: first-build timing, pre-build rejection count, camera conflict count, and same-context repetition discrimination.

No telemetry transport, persistence, RAF, timer, Babylon scene, Worker, AudioContext, global listener, gameplay mutation, or player-identifying data is introduced.

## Scope

Relative to #115 head `a61e77673b0a510807342face455917a935f450c`:

- 3 commits ahead / 0 behind;
- 2 added files;
- no network/analytics backend;
- no affordability change;
- no input/camera behavior change;
- no balance constants.

## Validation

Post-freeze draft. Later root + Storybook gate:

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Additional pure fixtures should cover empty sessions, no successful build, non-monotonic timestamps, all rejection reasons, replacements with missing cell/role identity, changed-context replacements, and malformed numeric state.

## Follow-up

After #95 characterization and #115 certification, a small runtime adapter can record these events locally during certification/user-study fixtures. Keep the observer downstream of authoritative gameplay outcomes; it must never become a decision or economy owner.

This is post-freeze work and must not alter #92's active exact-SHA certification campaign.

### DIFF
--- a/experiments/storybook/src/Foundations/Interaction/DirectManipulationObservation.stories.ts
+++ b/experiments/storybook/src/Foundations/Interaction/DirectManipulationObservation.stories.ts
@@ -1,0 +1,222 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	classifyDirectManipulation,
+	type DirectManipulationCandidate
+} from "@defend/gameplay/directManipulation";
+import {
+	createDirectManipulationObservationState,
+	recordDirectManipulationObservation,
+	summarizeDirectManipulationObservations,
+	type DirectManipulationInterventionKind,
+	type DirectManipulationObservation,
+	type DirectManipulationObservationSummary
+} from "@defend/gameplay/directManipulationObservation";
+import { createLabShell } from "../../labTheme";
+
+function candidate(
+	overrides: Partial<DirectManipulationCandidate> = {}
+): DirectManipulationCandidate {
+	return {
+		targetKind: "ground",
+		gestureOwner: "world",
+		targetStale: false,
+		terrainValid: true,
+		occupied: false,
+		protected: false,
+		affordable: true,
+		requiredEnergy: 3000,
+		currentTowerLevel: null,
+		maximumTowerLevel: 3,
+		requestedTowerAction: "upgrade",
+		maintenanceAvailable: false,
+		...overrides
+	};
+}
+
+function observation(
+	atSeconds: number,
+	input: DirectManipulationCandidate,
+	interventionKind: DirectManipulationInterventionKind,
+	cellKey: string | null,
+	roleKey: string | null,
+	tacticalContextKey: string
+): DirectManipulationObservation {
+	return {
+		atSeconds,
+		cellKey,
+		roleKey,
+		tacticalContextKey,
+		interventionKind,
+		outcome: classifyDirectManipulation(input)
+	};
+}
+
+function summarizeSession(
+	events: DirectManipulationObservation[]
+): DirectManipulationObservationSummary {
+	let state = createDirectManipulationObservationState(0);
+	for (let index = 0; index < events.length; index += 1) {
+		state = recordDirectManipulationObservation(state, events[index]);
+	}
+	return summarizeDirectManipulationObservations(state);
+}
+
+function legibleSession(): DirectManipulationObservationSummary {
+	return summarizeSession([
+		observation(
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Rebuild guidance now that parent #115 is closed in favor of #112 + #117:

1. Retarget the observer to the canonical `TowerInteractionPreview`/reason vocabulary from #117 rather than copying #115's outcome types. Include the new `invalid-cost`, plus `outside-arena` and `invalid-tower-level`; do not preserve superseded names solely for compatibility with the closed parent.

2. Preserve #117's `allowed / rejected / ignored` distinction in the metrics. In particular, `camera-gesture` is an **ignored world action because camera arbitration won**, not a rejected placement. It should increment a dedicated camera-conflict/ignored count, but should not inflate `rejectedActions`, rejection attribution, or `rejectedBeforeFirstBuild`. Otherwise a user correctly orbiting the camera can make first-build friction look worse than it is.

3. Keep two pre-first-build measures if useful: true rejected world attempts, and camera arbitration conflicts. #108 explicitly wants both, and separating them makes it possible to tell economic/spatial comprehension problems from gesture ownership problems.

4. The current monotonic timestamp clamp prevents time moving backward but can move a late-arriving earlier success **forward** to the most recently observed time. For user-study timing, either reject/flag out-of-order samples or keep the original validated event timestamp while maintaining a separate monotonic ingestion cursor; do not silently rewrite event time if exact time-to-first-build is evidence.

The same-context replacement logic is otherwise a useful downstream measurement seam. Keep it observational only and outside frozen #92 heads.

xtreemze:
Superseded by #120 after concurrency review.

#116 was stacked on duplicate classifier #115. Concurrent #112 + #117 are now the authoritative interaction chain, with explicit legacy/inclusive affordability policy, max-tier policy, outside-arena/invalid-level classification, and malformed-cost fail-closed handling.

#120 ports the useful passive observation semantics onto #117 and improves camera accounting: camera-owned gestures remain `ignored` rather than being counted as gameplay rejections, while still surfacing as first-build arbitration conflicts.

No production code depends on this branch.

PR #119, opened 2026-09-04
experiment: make raid commitment respond to perceived target value

xtreemze:
Refs #107, #103
Parent: #114
Related: #111, #74, #76, #86

## Purpose

Move target richness one step earlier in the deterrence causal chain: before a raider is committed, estimate whether the perceived target contains enough recoverable value to justify that tier's embodied/launch energy and travel/operating cost.

#114 corrects conserved settlement **after** a raid reaches the target. This child adds the complementary pre-commit estimate so a poor target can reduce expensive commitment before the breach is attempted.

## `src/js/gameplay/raidExpectedValue.ts`

Adds an engine-free expected-value model with explicit inputs:

- raider tier;
- committed/embodied energy;
- extraction potential;
- **perceived target energy**;
- expected breach probability;
- expected arrival viability;
- travel/operating cost.

It derives:

- viability-limited extraction potential on arrival;
- target-limited extraction if a breach succeeds;
- probability-weighted expected extraction;
- expected net return;
- expected return on commitment.

### Perception boundary

`perceivedTargetEnergy` is intentionally not authoritative silo state. It may later come from the lossy brightness/throughput/reachability signature discussed in #107/#103.

The important invariant is economic: even a weakly defended target should not appear capable of yielding more energy than the attacker believes is present.

Authoritative after-the-fact transfer remains #114's responsibility.

## Admission / diagnostic selection

`raidExpectedValueIsAdmissible()` accepts a caller-supplied minimum expected-net threshold.

- zero/positive can model strict economic discipline;
- a modest negative threshold can model deliberate scouting/probing under uncertainty.

`selectRaidTierByExpectedNetReturn()` is explicitly a **diagnostic experiment helper**, not final AI. Among admissible options it picks highest expected absolute net, with ties favoring lower committed energy so tier number itself never wins a tie.

#107 should still compare multiple commitment/probe policies rather than making this selector canonical.

## Storybook playground

Adds `Foundations/Strategy/Raid Expected Value`.

It compares three ordered perceived target values across R1/R2/R3 with controls for breach probability, arrival viability, travel cost, target values and admission threshold.

The fixed deterministic witness uses current economy-hypothesis scales:

- R1: commit 3000 / extraction potential 7720;
- R2: 12000 / 30440;
- R3: 27000 / 68160;
- expected breach 0.96;
- expected arrival viability 0.95;
- travel/operating cost 500.

Under that witness:

- perceived 30000 → diagnostic selector favors R2;
- perceived 5000 → only R1 remains positive-EV;
- perceived 1000 → no tier is positive-EV.

This is **not balance certification**. It is a structural witness that target richness can naturally cause `expensive commitment → probe → no economically rational raid` as value disappears.

The story also asserts expected extracted energy never exceeds perceived target energy × breach probability.

No RAF, timer, Babylon scene, Worker, AudioContext or production AI is introduced.

## Scope

Relative to #114 head `11fa4384bd3a3e5c8d475a9da0311918d8d0f6cd`:

- 2 commits ahead / 0 behind;
- 2 added files;
- no edits to #111's current scenario runner;
- no production wave/AI changes;
- no final confidence, launch-cost or target-signature constants.

## Integration into #111

After #114 conservation semantics are adopted, #111 should replace state-only raid frequency/tier selection with a model that also consumes perceived target value and expected physical outcomes.

At minimum test:

1. rich + vulnerable target can restore attacker interest;
2. poor target suppresses expensive bodies even if vulnerable;
3. cheap probes persist longer under uncertainty than expensive bodies;
4. repeated unprofitable probes eventually go quiet;
5. one lucky probe does not instantly restore maximum commitment;
6. no tier gets a hidden ROI bonus from tier number.

## Validation

Post-freeze draft. Later root + Storybook gate:

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Add sweeps across target energy, breach probability, viability, travel cost and admission threshold. Compare against measured physical outcomes once #72/#86 fixtures are available.

This does not alter #92's frozen certification campaign.

### DIFF
--- a/experiments/storybook/src/Foundations/Strategy/RaidExpectedValue.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/RaidExpectedValue.stories.ts
@@ -1,0 +1,194 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	estimateRaidExpectedValue,
+	raidExpectedValueIsAdmissible,
+	selectRaidTierByExpectedNetReturn,
+	type RaidExpectedValueEstimate,
+	type RaidExpectedValueTier
+} from "@defend/gameplay/raidExpectedValue";
+import { createLabShell } from "../../labTheme";
+
+type ExpectedValueArgs = {
+	richTargetEnergy: number;
+	poorTargetEnergy: number;
+	emptyTargetEnergy: number;
+	expectedBreachProbability: number;
+	expectedArrivalViability: number;
+	travelAndOperatingCost: number;
+	minimumExpectedNetReturn: number;
+};
+
+type TierProfile = {
+	tier: RaidExpectedValueTier;
+	committedEnergy: number;
+	extractionPotential: number;
+};
+
+const PROFILES: TierProfile[] = [
+	{ tier: 1, committedEnergy: 3000, extractionPotential: 7720 },
+	{ tier: 2, committedEnergy: 12000, extractionPotential: 30440 },
+	{ tier: 3, committedEnergy: 27000, extractionPotential: 68160 }
+];
+
+function estimatesFor(
+	perceivedTargetEnergy: number,
+	args: ExpectedValueArgs
+): RaidExpectedValueEstimate[] {
+	return PROFILES.map(profile =>
+		estimateRaidExpectedValue({
+			tier: profile.tier,
+			committedEnergy: profile.committedEnergy,
+			extractionPotential: profile.extractionPotential,
+			perceivedTargetEnergy,
+			expectedBreachProbability: args.expectedBreachProbability,
+			expectedArrivalViability: args.expectedArrivalViability,
+			travelAndOperatingCost: args.travelAndOperatingCost
+		})
+	);
+}
+
+function fixed(value: number, digits = 0): string {
+	if (value !== value || value === Infinity || value === -Infinity) return "0";
+	return value.toFixed(digits);
+}
+
+function targetCard(
+	label: string,
+	targetEnergy: number,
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
This pre-commit EV seam correctly target-bounds expected extraction, but it currently inherits #111's attacker-side conservation problem: `expectedNetReturn = expectedExtractedEnergy - committedEnergy - travelAndOperatingCost` treats 100% of embodied launch energy as irrecoverably spent even when the modeled raid is expected to arrive with substantial viability. Under #79, surviving raiders evacuate and can reintegrate into the mothership reserve, so commitment should be separated from permanent embodied loss.

Before this becomes the tier-admission foundation, add an explicit expected returned/surviving embodied term (preferably caller-supplied rather than assuming HP/viability maps 1:1 to teal energy). Conserved form: `expected net = expected extraction + expected returned embodied - committed embodied - travel/operating cost`, equivalently `expected extraction - expected permanently lost embodied - travel/operating cost`. The return term should account for outcome probability/recovery semantics and then reconcile with #114's capture-eligible partition after settlement.

This matters to the default selector witness: durable R2/R3 commitments may be structurally undervalued today, so `rich -> R2 / medium -> R1 / poor -> none` should not be treated as meaningful even experimentally until the attacker embodied-energy lifecycle is closed. I recorded the corresponding settlement invariant/ROI requirement on #107 and #111.

Secondary robustness point: malformed/negative `committedEnergy` is sanitized to zero while extraction inputs can remain positive, which can fabricate a free positive-EV candidate in diagnostics. Consider explicit input validity/fail-closed admission rather than allowing invalid economic profiles to participate in selection.

PR #120, opened 2026-09-04
Add passive tower-interaction comprehension metrics

xtreemze:
Refs #108, #103
Parent chain: #112 → #117
Related: #92, #95, #109

## Purpose

Provide the passive evidence layer for #108's first-build discoverability, rejection attribution, camera/placement arbitration, and maintenance-repetition gates while reusing the authoritative concurrent interaction classifier instead of maintaining a duplicate seam.

This replaces/supersedes #116, which was originally stacked on the now-closed duplicate classifier #115.

## `src/js/gameplay/towerInteractionObservation.ts`

Consumes #117's `TowerInteractionPreview` downstream of classification. It never changes intent, disposition, reason, cost, affordability or tower state.

Each observation supplies:

- session-relative timestamp;
- optional stable cell identity;
- optional semantic role identity;
- caller-owned coarse tactical-context fingerprint;
- intervention kind (`new-build / replacement / upgrade / maintenance / other`);
- authoritative interaction preview.

The observer derives:

- total attempts;
- allowed actions;
- rejected actions;
- ignored actions;
- allowance rate;
- rejection counts for occupied, unaffordable, invalid-cost, protected-core, invalid-terrain, outside-arena, stale-target, max-level and invalid-level states;
- camera-gesture count;
- rejected attempts before first build;
- ignored attempts before first build;
- time to first allowed action;
- time to first successful build;
- accepted replacements;
- same-cell/same-role replacements under unchanged tactical context;
- repetition rate.

### Camera semantics

#112/#117 classify camera-owned gestures as `ignored`, not rejected. This observer preserves that distinction. A camera conflict still counts as an attempt and is surfaced separately for the first-build gate, but does not inflate gameplay rejection counts.

### Maintenance/input-tax boundary

`tacticalContextKey` is caller-owned and coarse. Same-cell/same-role replacement is counted as rote only if the last accepted context fingerprint for that cell+role is unchanged.

A replacement after a material change in enemy mix, topology, reserve pressure, breach state or another relevant factor should receive a changed fingerprint and therefore not be mislabeled as busywork.

## Storybook playground

Adds `Arena/Interaction/Tower Observation`.

Two deterministic evidence-shape fixtures are shown:

### Legible first contact
- occupied rejection;
- camera-owned ignored gesture;
- first build at 4.2 s;
- two replacements, one before and one after tactical context changes;
- later upgrade.

### High-friction / rote maintenance
- occupied/protected/resource rejections;
- camera conflict;
- later first build;
- three same-cell/same-role replacements under unchanged context.

The story verifies:

- camera ignored vs rejection separation;
- first-build timing;
- rejected/ignored attempts before first build;
- camera conflict counting;
- unchanged-context replacement discrimination.

The illustrative numbers are not UX targets.

No analytics transport, persistence, player-identifying information, RAF, timer, Babylon scene, Worker, AudioContext, global listener or gameplay mutation is introduced.

## Scope

Relative to #117 head `0de9606e2bbdf1e917a28bca3a1e0487520cece0`:

- 2 commits ahead / 0 behind;
- 2 added files;
- no input/economy/tower mutation;
- no frozen certification-head changes.

## Validation

Post-freeze draft. Later root + Storybook campaign should cover empty sessions, no successful build, non-monotonic timestamps, every #117 reason, ignored camera attempts, replacements with missing identities, changed-context replacements, malformed observer state, plus:

```sh
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

## Follow-up

After #95 characterization and the #112/#117 chain is certified, a minimal Babylon adapter may record local certification/user-study observations. Keep this observer downstream of authoritative interaction decisions; metrics must never feed back into gameplay authority.

This remains outside the frozen #92 campaign.

### DIFF
--- a/experiments/storybook/src/Arena/Interaction/TowerInteractionObservation.stories.ts
+++ b/experiments/storybook/src/Arena/Interaction/TowerInteractionObservation.stories.ts
@@ -1,0 +1,229 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	classifyTowerInteraction,
+	type TowerInteractionRequest
+} from "@defend/gameplay/towerInteraction";
+import {
+	createTowerInteractionObservationState,
+	recordTowerInteractionObservation,
+	summarizeTowerInteractionObservations,
+	type TowerInteractionInterventionKind,
+	type TowerInteractionObservation,
+	type TowerInteractionObservationSummary
+} from "@defend/gameplay/towerInteractionObservation";
+import { createLabShell } from "../../labTheme";
+
+function request(
+	overrides: Partial<TowerInteractionRequest> = {}
+): TowerInteractionRequest {
+	return {
+		inputOwner: "world",
+		targetKind: "ground",
+		occupied: false,
+		balance: 10000,
+		requestedCost: 3000,
+		currentLevel: 0,
+		maximumLevel: 3,
+		affordabilityRule: "legacy-strict",
+		maxLevelBehavior: "no-op",
+		...overrides
+	};
+}
+
+function observation(
+	atSeconds: number,
+	input: TowerInteractionRequest,
+	interventionKind: TowerInteractionInterventionKind,
+	cellKey: string | null,
+	roleKey: string | null,
+	tacticalContextKey: string
+): TowerInteractionObservation {
+	return {
+		atSeconds,
+		cellKey,
+		roleKey,
+		tacticalContextKey,
+		interventionKind,
+		preview: classifyTowerInteraction(input)
+	};
+}
+
+function summarizeSession(
+	events: TowerInteractionObservation[]
+): TowerInteractionObservationSummary {
+	let state = createTowerInteractionObservationState(0);
+	for (let index = 0; index < events.length; index += 1) {
+		state = recordTowerInteractionObservation(state, events[index]);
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Static evidence-integrity review found two follow-ups before certification:

1. `recordTowerInteractionObservation()` currently handles a non-monotonic `observation.atSeconds` by taking `max(lastObservedAtSeconds, observation.atSeconds)` and then using that coerced timestamp for the event. This can silently reorder an earlier event into the future and affect first-build timing/context sequence. The PR's validation section explicitly calls out non-monotonic timestamps; prefer fail-closed/explicit diagnostic semantics rather than silently changing event time.

2. The validation section also calls out malformed observer state, but `copyReasonCounts(state.reasonCounts)` and `copyContexts(state.lastAcceptedContexts)` assume structurally valid state, and numeric reason counters are copied without finite/non-negative sanitization. Corrupt/rehydrated state can therefore throw or propagate NaN/Infinity into summaries. Consider sanitizing state at the observer boundary and exposing invalid/non-monotonic observation counts, keeping malformed evidence out of first-build/repetition rates.

These are diagnostic integrity issues only; they do not require any gameplay-authority change. #129 uses #120 downstream but does not modify its files, so the fixes can remain owned here or in a narrow child.

PR #123, opened 2026-09-04
Add deterministic defense interaction legibility contract

xtreemze:
Refs #108
Related: #33, #56, #92, #109

## Purpose

Create the first implementation seam for the ludic/intuitiveness work without changing live placement, upgrade, balance, camera, or Babylon behavior.

The core problem from #108 is that direct-world interaction is mechanically simple but rejected/ambiguous actions are not represented by an explicit semantic contract. This PR makes those outcomes classifiable before presentation adapters are added.

## Pure interaction contract

Adds `src/js/gameplay/defenseInteraction.ts`.

The Babylon/DOM-independent resolver distinguishes:

- accepted placement;
- accepted tower upgrade;
- explicit max-level refresh when a caller intentionally selects that policy;
- camera-owned gesture;
- stale/missing target;
- invalid terrain;
- protected/core target;
- occupied target;
- unaffordable action;
- invalid tower state;
- max semantic tower state.

It returns an explicit action (`place`, `upgrade`, `refresh`, `none`), rejection reason, and tower before/after levels where relevant.

### Important affordability boundary

This module deliberately receives `affordable: boolean` from the caller. It does **not** encode the current legacy `balance > cost` rule or #109's intended `balance >= cost` correction.

That keeps #109 independently fixable after the baseline is characterized and prevents this PR from contaminating #95/#97 certification semantics.

## Storybook playground

Adds `Foundations/Interaction/Defense Interaction Legibility`.

The deterministic matrix now covers every outcome in the resolver contract:

- valid placement;
- invalid terrain;
- protected core;
- occupied cell;
- insufficient energy;
- camera gesture ownership;
- stale target;
- normal tower upgrade;
- unaffordable tower upgrade;
- max-state rejection;
- explicit max-state refresh policy;
- invalid tower semantic state.

Accepted/rejected state is conveyed by text as well as presentation styling, so the fixture does not depend on color alone. The `play` assertion iterates the complete matrix, verifies each semantic action/rejection pair, clicks each scenario, and checks the live result announcement.

The story creates no Babylon scene, RAF loop, Worker, AudioContext, timer, or production event listener.

## Scope

Current head: `13e0b5ee7f5135f2411b202d19bd1791699b3119`.

Relative to current `master`:

- 3 commits ahead / 0 behind;
- exactly 2 added files;
- no live production call-site changes;
- no balance/cost change;
- no pointer/camera behavior change;
- no dependency/package change;
- no #56 hex-topology ownership;
- no #95/#97/#105 modification.

## Certification

Keep draft until a later local certification campaign can run:

```sh
npm run lint
npm run build
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Then exercise the story with keyboard/pointer input and confirm all rejected cases expose a distinct reason without relying only on color.

The root gate should specifically confirm the helper remains compatible with the historical TypeScript target.

## Follow-up after certification

1. adapt current pointer/touch targeting to produce this semantic request/result without changing behavior first;
2. use the result to drive world-grounded feedforward and rejection feedback;
3. implement #109's affordability boundary as a separate narrow change;
4. build the deterministic first-contact comprehension fixture from #108.

This PR is intentionally a semantic/testing foundation, not the final presentation implementation.

### DIFF
--- a/experiments/storybook/src/Foundations/Interaction/DefenseInteraction.stories.ts
+++ b/experiments/storybook/src/Foundations/Interaction/DefenseInteraction.stories.ts
@@ -2,6 +2,8 @@ import type { Meta, StoryObj } from "@storybook/html-vite";
 import { expect, userEvent } from "storybook/test";
 import {
   resolveDefenseInteraction,
+  type DefenseInteractionAction,
+  type DefenseInteractionRejection,
   type DefenseInteractionRequest,
   type DefenseInteractionResolution
 } from "@defend/gameplay/defenseInteraction";
@@ -12,6 +14,8 @@ type Scenario = {
   title: string;
   note: string;
   request: DefenseInteractionRequest;
+  expectedAction: DefenseInteractionAction;
+  expectedRejection: DefenseInteractionRejection;
 };
 
 const scenarios: Scenario[] = [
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Static/architectural review at exact head `13e0b5ee7f5135f2411b202d19bd1791699b3119`: scope remains exactly two added files with no live call-site, balance, Babylon, camera, dependency, or topology changes. The resolver keeps affordability external, so #109 remains independently fixable after baseline characterization. The Storybook fixture now covers every declared rejection reason plus accepted placement/upgrade/refresh outcomes and exposes accepted/rejected state textually rather than by color alone. No static blocker found. Keep draft: historical root TypeScript/build and Storybook typecheck/build/test/browser evidence are still required before promotion.

xtreemze:
Superseded after concurrent-state review by the stronger existing chain #112 → #117 → #120.

#112 already owns a richer interaction contract (cost/balance projections, outside-arena state, camera `ignored` semantics, explicit legacy/inclusive affordability policy, and explicit max-tier policy). #117 hardens malformed cost handling fail-closed, and #120 consumes that authoritative preview for first-build/rejection/maintenance observations.

Maintaining `defenseInteraction.ts` in parallel would create two semantic authorities for the same pointer/tower action and force later adapters to choose between overlapping result vocabularies. The useful ideas from this PR are already represented in the concurrent chain, so this PR should not be certified or merged.

No frozen #95/#97/#105 head is affected.

Issue #124, opened 2026-09-04
Test raider launch energy as recoverable embodied capital


xtreemze:
Parent: #107
Related: #74, #76, #103, #111, #114, #119

## Purpose

Resolve a conservation/economic contradiction exposed when #114/#119 make target reserve authoritative.

The current experimental raid hypothesis treats `launchCost = 3000 × tier²` as fully sunk at launch while extraction is capped by energy physically present in the defender silo.

With the current 30,000 defender reserve cap this breaks the earlier claim that R1/R2/R3 have roughly symmetric economic break-even points.

## The contradiction

Current hypotheses:

- R1 launch commitment: 3,000; extraction potential: 7,720;
- R2 launch commitment: 12,000; extraction potential: 30,440;
- R3 launch commitment: 27,000; extraction potential: 68,160;
- target reserve cap: 30,000;
- representative travel/operating cost in #111/#119: 500.

Once extraction is conserved against the target reserve, maximum possible net return from a completely full target is approximately:

- R1: `min(7,720, 30,000) - 3,000 - 500 = +4,220`;
- R2: `min(30,440, 30,000) - 12,000 - 500 = +17,500`;
- R3: `min(68,160, 30,000) - 27,000 - 500 = +2,500`.

For R3 to have positive *expected* value when requested extraction is already target-capped at 30,000, its probability-weighted realized extraction must exceed 27,500. That implies nearly certain success (`>91.7%` effective extraction probability) even against a full target.

Therefore the earlier ~40% remaining-health break-even calculation only holds when target supply is effectively uncapped. It does not hold under the canonical 30,000 conserved silo reserve.

This is a structural issue, not parameter noise.

## Primary refinement hypothesis: launch energy is capital at risk

Treat raider launch energy as **embodied capital transferred from mothership reserve into the raider body**, not energy destroyed at launch.

Lifecycle:

`mothership reserve → embodied raider energy → damage/loss/capture OR surviving return → mothership reserve`

A surviving raider returns some or all remaining embodied energy when it evacuates. The attacker's true economic loss is therefore primarily the damaged/destroyed/non-returning fraction plus travel/operating drain.

Conceptually:

`returnedCapital = launchCommitment × returnableViability`

`lostCapital = launchCommitment - returnedCapital`

`attackerNet = actualExtractedEnergy - lostCapital - travel/operatingCost`

This is an experimental equation, not a production constant.

### 1. Conservation becomes literal

[...]

PR #125, opened 2026-09-04
experiment: treat raider launch energy as capital at risk

xtreemze:
Refs #124, #107, #103
Parent: #119
Related: #114, #74, #76

## Purpose

Compare the current fully-sunk launch-cost interpretation with a conservation-first hypothesis where launch energy moves into the raider as embodied capital and returns only in proportion to physical survival.

This addresses the structural problem documented in #124: once target extraction is correctly capped by the canonical 30,000 silo reserve, a fully sunk 27,000 R3 commitment has positive expected value only under near-certain rich-target extraction. The earlier tier-symmetric ~40% break-even argument therefore does not survive conserved target supply.

## Scope

Relative to #119 exact head `720b3fa880a8d7fd388f33baa6b2e52497f6487f`:

- 3 commits ahead / 0 behind;
- exactly 2 added files;
- `src/js/gameplay/raiderCapitalEconomy.ts`;
- `experiments/storybook/src/Foundations/Strategy/RaiderCapitalEconomy.stories.ts`;
- no final launch-cost, return-efficiency or balance promotion;
- no changes to the frozen #92 certification heads.

The branch inherits #114 conserved target settlement and #119 perceived-target expected-value work.

### Fully sunk baseline

Launch commitment is an explicit energy sink at deployment. It cannot simultaneously be defender-capturable embodied energy without introducing another source.

### Embodied return hypothesis

`mothership reserve -> raider embodied capital -> surviving return OR damage/capture/loss`

For the first experiment:

`returnedCapital = commitment × remainingViability`

`capitalLoss = commitment - returnedCapital`

`attackerNet = actualTargetBoundedExtraction - capitalLoss - travel/operatingCost`

This is a hypothesis, not production semantics.

## Conserved realized settlement

`settleRaiderCapitalEconomy()` composes #114's `settleConservedRaidExchange()`.

For embodied capital:

- extraction is capped by actual defender reserve;
- surviving capital returns according to physical viability;
- only non-returning capital is exposed to defender capture;
- defender capture is bounded by that damaged/lost capital and storage headroom;
- uncaptured non-returning capital remains explicitly loose/dissipative for later physical interpretation;
- collateral loss remains dissipative rather than attacker income;
- an explicit conservation residual verifies the tracked target + launch-capital exchange reconciles.

For the sunk baseline, commitment is tracked as an explicit launch sink and cannot also produce defender recovery.

Travel/operating drain is reported in attacker net but intentionally outside the per-contact tracked-energy identity because it is a separate mothership/system sink.

## Expected-value comparison

`estimateRaiderCapitalExpectedValue()` extends the #119 target-richness idea with expected returned capital.

Inputs remain explicit and injectable:

- perceived target energy;
- breach probability;
- arrival viability on breach;
- return viability after a failed breach/retreat;
- travel/operating cost.

Expected extraction remains target-bounded.

## Storybook witness

Adds `Foundations/Strategy/Raider Capital Economy`.

The default moderate-risk witness uses:

- breach probability 0.55;
- successful arrival/return viability 0.55;
- failed-breach return viability 0.30;
- travel/operating cost 500;
- target reserves 5k / 15k / 30k.

Under the **embodied-capital** hypothesis, the fixed witness intentionally demonstrates contextual economic viability rather than universal tier efficiency:

- 5k poor target: R1 positive, R2/R3 negative;
- 15k medium target: R2 positive, R3 negative;
- 30k rich target: R3 becomes positive.

The exact values are not balance targets. The point is that target richness, physical survival and concentrated capital risk can produce tier differentiation without a hidden tier-specific ROI bonus.

Controls expose probability, success/failure viability, realized viability, operating cost and defender capture fraction.

## Automated invariants

The Storybook fixture verifies across both capital modes, three target reserves and all three tiers:

- actual extraction never exceeds target reserve;
- returned capital never exceeds committed energy;
- defender-captured capital never exceeds non-returning embodied capital;
- the tracked target + launch-capital exchange has near-zero conservation residual;
- fully sunk launch cost returns no launch capital and yields no capture from that sunk source;
- the default embodied witness has the intended poor→medium→rich tier-viability progression.

## Design interpretation

This model strengthens the existing `physics <-> economy` pillar:

- a surviving raider preserves capital;
- damage destroys/exposes capital;
- defender recovery becomes captured opposing investment;
- ejection or failed evacuation can become real capital loss;
- R3 is risky because a large amount of capital is physically concentrated in one body, not because the game silently deletes a large tier tax.

It also makes intentional under-defense less attractive in a physically grounded way: weak defense lets raiders retain more of their own capital, gives the defender less capture, and allows more extraction.

## Validation / decision gate

Keep draft. Before choosing a capital model:

- run Storybook typecheck/build/test in a post-freeze campaign;
- sweep breach probability, target richness and success/failure viability;
- compare R1/R2/R3 contextual usefulness;
- add ejection/off-map and evacuation-return variants;
- feed measured viability distributions from #72/#86 rather than inventing permanent tier bonuses;
- compare long-horizon anti-farming behavior in #107;
- decide whether viability should map 1:1 to economic capital or through a separate bounded integrity curve.

Do not complete/promote the deterrence economy around fully sunk tier costs until this comparison is resolved.

### DIFF
--- a/experiments/storybook/src/Foundations/Strategy/RaiderCapitalEconomy.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/RaiderCapitalEconomy.stories.ts
@@ -1,0 +1,208 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	DEFAULT_RAIDER_CAPITAL_PROFILES,
+	estimateRaiderCapitalExpectedValue,
+	raiderCapitalExpectedValueIsPositive,
+	settleRaiderCapitalEconomy,
+	type RaiderCapitalMode,
+	type RaiderCapitalProfile
+} from "@defend/gameplay/raiderCapitalEconomy";
+import { createLabShell } from "../../labTheme";
+
+type CapitalEconomyArgs = {
+	expectedBreachProbability: number;
+	expectedBreachViability: number;
+	expectedFailureReturnViability: number;
+	realizedViability: number;
+	travelAndOperatingCost: number;
+	defenderCaptureFraction: number;
+};
+
+type TargetWitness = {
+	label: string;
+	reserve: number;
+};
+
+const TARGETS: TargetWitness[] = [
+	{ label: "Poor target", reserve: 5000 },
+	{ label: "Medium target", reserve: 15000 },
+	{ label: "Rich target", reserve: 30000 }
+];
+
+const MODES: RaiderCapitalMode[] = ["fully-sunk", "embodied-return"];
+
+function fixed(value: number, digits = 0): string {
+	if (value !== value || value === Infinity || value === -Infinity) return "0";
+	return value.toFixed(digits);
+}
+
+function signed(value: number): string {
+	return `${value >= 0 ? "+" : ""}${fixed(value, 0)}`;
+}
+
+function modeLabel(mode: RaiderCapitalMode): string {
+	return mode === "fully-sunk" ? "Fully sunk launch" : "Embodied capital return";
+}
+
+function estimateCell(
+	mode: RaiderCapitalMode,
+	profile: RaiderCapitalProfile,
+	target: TargetWitness,
+	args: CapitalEconomyArgs
+): string {
+	const estimate = estimateRaiderCapitalExpectedValue({
+		mode,
+		profile,
+		perceivedTargetEnergy: target.reserve,
+		expectedBreachProbability: args.expectedBreachProbability,
+		expectedArrivalViabilityOnBreach: args.expectedBreachViability,
+		expectedReturnViabilityOnFailure: args.expectedFailureReturnViability,
+		travelAndOperatingCost: args.travelAndOperatingCost
+	});
+	const positive = raiderCapitalExpectedValueIsPositive(estimate);
+	return `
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
Static capital-model review: conservation accounting is coherent, but the current viability contract collapses two physically distinct phases that matter to the design.

`remainingViability` in realized settlement both limits extraction (`extractionPotential * viability`) and determines returned capital (`commitment * viability`). The expected-value path does the same with `expectedArrivalViabilityOnBreach`: it limits extraction and is then treated as successful-breach return viability. This makes one important outcome impossible to represent: **reach the silo / extract successfully, then lose additional capital during evacuation** through interception, ejection, collision, or a failed return to the mothership.

Before this becomes the capital-at-risk foundation, split at least:

- viability/capability at extraction or breach;
- viability/capital actually returned after the post-breach escape/evacuation phase.

Then preserve identities such as `returnedCapital <= capital present after extraction` and expose post-breach capital loss separately. Add witnesses for (a) successful extraction + clean return, (b) successful extraction + severe return loss/ejection, and (c) failed breach + partial retreat survival. This is strategically important because otherwise defender action after silo contact has no economic leverage over already-successful raiders.

Separate robustness note for later hardening: the local `finite()` helper still assumes runtime numeric types, so `undefined as unknown as number` would propagate through `positive()` as NaN. If malformed/rehydrated state is in certification scope, apply the same runtime-number guard now used by #134/#141.

Issue #126, opened 2026-09-04
Model attacker beliefs so deterrence emerges from learning, not omniscience


xtreemze:
Parent: #107
Related: #73, #76, #103, #119, #124, #125

## Purpose

Make attacker adaptation and deterrence arise from **imperfect, history-updated beliefs** about a target rather than direct access to authoritative defense state.

The deterrence arc should be rational without being omniscient:

`attack → observe outcome → update belief → adapt composition → probe → quiet → information becomes stale → occasional re-probe`

This is necessary for both game design and campaign pacing.

## Problem exposed by the current experiments

Once #119 evaluates raids against target-bounded expected value, feeding it authoritative/current breach probability or current fortress effectiveness can make a competent defense immediately appear negative-EV.

An omniscient attacker then withdraws at once:

`strong defense → instant expected-value rejection → no adaptation/probing history → immediate starvation`

That defeats the intended narrative sequence. The player should first experience attackers **learning** that the fortress has become expensive, not the world silently switching off because an AI read hidden state.

The opposite failure is also possible: if attacker confidence is merely a scripted campaign meter, deterrence stops being a consequence of play.

## Canonical information boundary

Attackers may know or infer only information plausibly available from the world and previous interactions.

Potential observations:

- lossy target energy/signature brightness from #76;
- recent target activity/throughput;
- their own committed tier/composition;
- whether a breach occurred;
- remaining viability / survivors;
- extraction actually obtained;
- time-to-silo / route length;
- bodies lost/ejected/expired;
- visible topology or route openings at observational resolution;
- travel distance / operating cost;
- age of the last reliable observation.

They should **not** read authoritative hidden tower DPS, exact future route solution, true breach probability, exact defender reserve, or a global difficulty score.

## Belief state

First experiment should keep the state compact and explicit:

- `perceivedTargetEnergy`;
- `expectedBreachProbability`;
- `expectedArrivalViability`;
- `uncertainty` / information staleness;
- direct observation count;
- opportunities/time since last direct observation.

The belief should be sufficient to feed #119/#125 expected-value models without owning those economic calculations.

### Passive signature observation

A visible teal signature may update perceived target energy faster than combat estimates, but remains lossy.

A brightening target can therefore regain interest even after prior deterrence; a dimming target naturally becomes less economically attractive.

### Direct raid observation

A raid supplies stronger evidence:

- breach success/failure updates expected breach probability;
- observed surviving viability updates expected arrival/return viability;
[...]

PR #127, opened 2026-09-04
experiment: make deterrence learn from attacker beliefs

xtreemze:
Refs #126, #107, #103
Related: #73, #76, #119, #124, #125

## Purpose

Make deterrence compatible with rational but non-omniscient attackers by introducing a small history-updated belief state that can feed later expected-value planners.

The campaign should not jump from strong defense directly to silence because AI reads authoritative fortress strength. Attackers should learn through outcomes, reduce commitment, probe while information remains uncertain, go quiet when failure becomes well established, and eventually re-check a world whose old information has become stale.

## Scope

Base: `master` `e9852845601606db97fc7a2931b69fd7d1af6d23`.

Exactly two added files:

- `src/js/gameplay/raidBelief.ts`;
- `experiments/storybook/src/Foundations/Strategy/RaidBelief.stories.ts`.

No production AI/wave call sites, package/dependency metadata, economy constants, rendering, Babylon/Cannon objects, timers, Workers, or frozen #92 certification heads are changed.

## Belief boundary

The pure model owns estimates only:

- perceived target energy;
- expected breach probability;
- expected arrival viability;
- uncertainty/staleness;
- direct observation count;
- time since direct observation.

It does not choose a raider tier, calculate final ROI, mutate authoritative reserve, or inspect hidden defense state.

Downstream experiments can feed these estimates into #119's target-bounded expected value or #125's embodied-capital expected value.

## Passive signature observation

`observeRaidSignature()` moves perceived target energy toward a lossy observed teal-energy signal with a time-domain half-life.

It deliberately does **not** reveal breach probability or defense quality. A brightening target can become more interesting without magically becoming easier.

The smoothing is time-based so a constant signal is equivalent at different observation cadences.

## Direct raid evidence

`observeRaidOutcome()` gradually updates:

- breach expectation toward observed breach/failure;
- arrival viability toward observed survivor viability;
- uncertainty downward because direct evidence is fresh.

The update is bounded and caller-calibrated. One lucky result cannot reset a history of failures; repeated contrary evidence is required.

## Staleness

`ageRaidBelief()` leaves mean breach/viability beliefs unchanged but increases uncertainty toward 1 with a time-domain half-life.

This represents a quiet world's capacity to change through tower degradation, topology changes and reserve movement without granting the attacker free information about the direction of that change.

## Information-need diagnostic

`raidBeliefInformationNeed()` returns a bounded diagnostic:

`uncertainty × sqrt(perceived target richness)`

It does not authorize a raid or add hidden economic return. A later planner may compare this information need with the bounded cost of an R1 probe.

The intended consequence is:

- dim + uncertain target: little reason to inspect;
- valuable + well-known bad target: little immediate reason to inspect;
- valuable + stale target: meaningful reason for a cheap probe.

## Storybook witness

Adds `Foundations/Strategy/Raid Belief` showing:

1. optimistic prior;
2. repeated failed raids;
3. one lucky success;
4. repeated success;
5. deterrent quiet;
6. target signature brightening;
7. fresh probe.

The fixture demonstrates:

- repeated failures gradually lower breach and viability expectations;
- one success only partially reverses the history;
- repeated success can genuinely restore expectations;
- quiet increases uncertainty without changing physical means;
- passive brightness raises perceived opportunity without changing defense estimates;
- a fresh direct probe reduces uncertainty again.

## Deterministic invariants

The Storybook `play` test verifies the learning/staleness relationships above and additionally compares equivalent 30/60/120 Hz aging and passive-signature sequences. Time-domain updates converge to the same state within tight tolerance.

Malformed/non-finite prior inputs are also required to normalize to finite bounded state.

## Design consequence

This creates an explicit separation between:

- **world truth** — authoritative simulation;
- **observable evidence** — signatures and raid outcomes;
- **attacker belief** — imperfect estimates;
- **planner economics** — #119/#125;
- **behavior** — later commitment/probe policy.

That separation is important for fairness and legibility: an attacker should adapt because of evidence the player can plausibly understand, not because an invisible director knows the correct answer.

## Validation / follow-up

Keep draft until a later post-freeze campaign can run root compatibility plus Storybook typecheck/build/test.

Then:

- sweep update rates and staleness half-life;
- test noisy/partial-reliability observations;
- connect belief outputs to #119/#125 in a headless planner experiment;
- assign explicit bounded information value to probes rather than a tier-specific bonus;
- prove one lucky probe cannot restore heavy commitment immediately;
- test long-horizon `adapt → probe → quiet → stale → re-probe` behavior in #107;
- calibrate observation evidence from measured #72/#86 outcomes.

No permanent player-facing `confidence` meter should be inferred from this diagnostic model.

### DIFF
--- a/experiments/storybook/src/Foundations/Strategy/RaidBelief.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/RaidBelief.stories.ts
@@ -38,12 +38,11 @@ function percent(value: number): string {
 
 function buildStages(args: RaidBeliefArgs): BeliefStage[] {
 	const calibration = DEFAULT_RAID_BELIEF_CALIBRATION;
-	const initial = createRaidBelief(DEFAULT_RAID_BELIEF_PRIOR, calibration);
-	let state = initial;
+	let state = createRaidBelief(DEFAULT_RAID_BELIEF_PRIOR, calibration);
 	const stages: BeliefStage[] = [
 		{
 			label: "Prior",
-			note: "Valuable target, optimistic but uncertain physical expectations.",
+			note: "One target-wide opportunity estimate plus one optimistic but uncertain approach context.",
 			state
 		}
 	];
@@ -57,7 +56,7 @@ function buildStages(args: RaidBeliefArgs): BeliefStage[] {
 	}
 	stages.push({
 		label: "Repeated failures",
-		note: "Direct evidence lowers breach/viability expectations and reduces uncertainty.",
+		note: "Direct evidence lowers this approach's breach/viability expectation without changing target-wide energy belief.",
 		state
 	});
 
@@ -68,27 +67,31 @@ function buildStages(args: RaidBeliefArgs): BeliefStage[] {
 	);
 	stages.push({
 		label: "One lucky success",
-		note: "Belief recovers partially; one result does not erase the accumulated history.",
+		note: "The contextual approach recovers partially; one result does not erase accumulated history.",
 		state
 	});
 
 	for (let index = 0; index < 4; index += 1) {
 		state = observeRaidOutcome(
 			state,
-			{ breached: true, remainingViability: Math.min(1, args.successViability + 0.08), reliability: 1 },
+			{
+				breached: true,
+				remainingViability: Math.min(1, args.successViability + 0.08),
+				reliability: 1
+			},
 			calibration
 		);
 	}
 	stages.push({
 		label: "Repeated success",
-		note: "Accumulated contrary evidence can restore broader confidence.",
+		note: "Accumulated contrary evidence can genuinely restore this approach's expectations.",
 		state
 	});
 
 	state = ageRaidBelief(state, args.quietSeconds, calibration);
 	stages.push({
 		label: "Deterrent quiet",
-		note: "Physical means remain unchanged, but old information becomes stale and uncertainty rises.",
+		note: "Approach means stay fixed, but old route/tier information becomes stale and uncertain.",
 		state
 	});
 
@@ -100,7 +103,7 @@ function buildStages(args: RaidBeliefArgs): BeliefStage[] {
 	);
 	stages.push({
 		label: "Target brightens",
-		note: "Passive teal evidence raises perceived opportunity without revealing hidden defense quality.",
+		note: "The shared target opportunity rises; the approach-specific breach estimate does not magically improve.",
 		state
 	});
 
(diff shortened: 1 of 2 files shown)


xtreemze:
Follow-up structural refinement is now on the same draft head (`f420580def7077f9b421b7d46e1e5a7aff46daab`). The belief contract no longer treats the fortress as one scalar physical estimate. It separates:

- target-wide `RaidOpportunityBelief` (perceived energy/signature); and
- contextual `RaidApproachBelief` (breach probability, arrival viability, uncertainty, evidence age).

Callers should share opportunity across a target while maintaining approach beliefs per materially different sector/tier/insertion context. This prevents one failed R1 corridor from teaching the AI that every R3 insertion is equally bad, while still allowing the target's global teal signature to affect all approaches. The Storybook fixture and cadence/malformed-input assertions were updated accordingly. Scope remains exactly two added files relative to the recorded base.

xtreemze (commented):
Belief-boundary review: the target/approach split is the right ownership model, but observation validity currently changes belief in ways that can mistake missing data for physical evidence.

1. **Passive signature failure becomes an empty target.** `observeRaidOpportunitySignature()` clamps `observedTargetEnergy`; the shared `clamp()` maps non-finite input to the minimum (`0`). With positive `deltaSeconds`, a NaN sensor reading therefore moves `perceivedTargetEnergy` toward zero rather than leaving the belief unchanged/marking the observation invalid. A sensor/data failure should not be evidence that the silo is empty.

2. **Zero-reliability direct evidence still refreshes history.** `observeRaidApproachOutcome()` with `reliability = 0` leaves means/uncertainty unchanged, but still increments `directObservations` and resets `secondsSinceDirectObservation` to zero. That reports fresh direct knowledge when the caller explicitly said the evidence carries no weight.

3. The same issue applies to malformed `remainingViability`: `clamp01(NaN)` becomes `0`, so a reliable-but-malformed result can teach catastrophic viability.

Before this feeds a planner, make evidence validity explicit. Recommended semantics: invalid passive observation => opportunity no-op (optionally diagnostic counter); reliability <= 0 => approach no-op and do not reset evidence age/count; invalid physical outcome with positive reliability => reject/ignore the observation rather than coerce it into failure. Add deterministic fixtures proving malformed/missing observations cannot move means, reduce/increase uncertainty, or refresh evidence age.

Runtime robustness should also use `typeof value === "number"` at these external evidence boundaries, consistent with #134/#141.

Issue #137, opened 2026-09-04
Make the world remember conflict through persistent physical residue


xtreemze:
Parent: #103
Related: #73, #76, #79, #89, #113

## Purpose

Strengthen Defend's systemic narrative without adding exposition by making prior conflict remain legible in the physical world.

The campaign already has the right ingredients in separate systems:

- #79: failed motherships become persistent hulks;
- #89: impacts, eruptions and construction can deform terrain;
- #76: energy can pool, strand, drain and leave depleted strategic signatures;
- #73: the original fortress remains as an abandoned relic during the defender → mothership inversion;
- #113: asymmetric victory can be rewarded partly through persistent wreckage/environmental history.

What is still missing is a cross-system design contract for **historical residue**: enough durable evidence that a player can enter or revisit a place and infer something about what happened there without a log, cutscene, or lore text.

The goal is not decoration. The world should become a partial causal record of prior play.

---

# Core principle

> Important systemic events should sometimes leave bounded, interpretable residue whose form follows from the event that created it.

A player should increasingly be able to look at a location and infer propositions such as:

- a very large commitment failed here;
- this defense once extended farther than it does now;
- this route was repeatedly contested;
- this source was heavily consumed or erupted;
- this fortress became deterrent and then starved;
- this target was recently extracted;
- this wreck belonged to the same mobile lifecycle state the player now inhabits.

This is **environmental causality**, not collectible lore.

---

# Residue classes

## 1. Mothership hulks

Preserve #79's principle that loss of lift leaves recognizable matter rather than a despawn.

Residue should preserve enough silhouette/structure to communicate:

- raider-family spherical geometry;
- approximate scale of the failed vessel;
- exhausted/dim core cavity;
- crash direction/impact relationship where readable;
- optional remaining recoverable teal matter.

A hulk may become bounded obstacle geometry when safe, but narrative persistence must not randomly invalidate later encounters.

## 2. Abandoned fortress remnants

The player's original fortress should not become an abstract completed level after conversion.

Where campaign/world continuity permits, preserve evidence such as:

- empty central foundation/core cavity after the silo transforms;
- degrading tower bases or lower-tier remnants;
- old barrier geometry;
- dry/depleted geothermal connections;
- prior impact deformation;
- stranded or slowly collected residual energy.

The first abandoned fortress is especially important because it is the player's first physical evidence that a silo and mothership are lifecycle states of one system.

## 3. Terrain memory

[...]

Issue #113, opened 2026-09-04
Design asymmetric victory encounters around physical and economic leverage


xtreemze:
Parent: #29
Related: #56, #72, #73, #79

## Purpose

Make **winning from an apparently inferior position** a deliberate recurring source of mastery in Defend.

The game already contains the correct systemic ingredients: finite-lived raiders, physical obstruction, mass and momentum, edge ejection, temporary defenses, shared energy economics, differentiated body sizes, and a raider mothership that loses energy while a raid remains unresolved. The encounter layer should intentionally create situations where the player appears outmatched in one dimension but can win by discovering leverage in another.

This must not become conventional difficulty through HP inflation or scripted heroic reversals. The player should win because the simulation made their decision structurally superior.

## Core principle

> Some of Defend's most memorable victories should begin by looking unfair. The player succeeds not by matching the opponent's strength, but by discovering which dimension of strength does not matter in the present geometry, timing, or economy.

Examples:

- numbers lose to congestion;
- mass loses to edge geometry;
- health loses to time;
- firepower loses to obstruction;
- expensive attackers lose to cheap redirection;
- range loses to occlusion;
- a locally weak defense wins because the attacker cannot sustain the raid economically;
- a small raiding force wins because insertion geometry bypasses most of the fortress's nominal strength.

### 1. Numerical asymmetry

The player faces more bodies than their apparent defensive capacity can destroy directly.

Correct response should often be:

- funnel;
- stagger arrival times;
- create collisions/congestion;
- let finite-life decay work;
- use a small number of decisive ejections rather than attempting complete attrition.

The important feeling is: **"I cannot kill all of these" → "I do not need to."**

### 2. Mass asymmetry

A large R2/R3 body appears to overpower local defenses.

Possible leverage:

- tangential Tower 3 impulse near an edge;
- repeated low-angle steering from T2 crossfire;
- terrain or barriers turning forward momentum into lateral displacement;
- forcing the body into geometry where its large diameter becomes a disadvantage.

Large bodies should sometimes be harder to stop but easier to exploit once their trajectory becomes unfavorable.

### 3. Economic asymmetry

The player spends far less energy than the opponent commits.

A 3,000-energy barrier that causes a very expensive raider to arrive below break-even viability should feel like a major victory even if the raider technically reaches the silo.

On the raider side, a cheap scout that exposes a weak approach and enables one profitable follow-up can outperform a brute-force expensive deployment.

[...]

PR #140, opened 2026-09-04
experiment: model bounded physical world residue

xtreemze:
Refs #137, #103
Related: #73, #76, #79, #89, #113, #92

## Purpose

Make #137's environmental-history idea executable without adding production persistence, Babylon objects, physics bodies, save-state migration, or narrative exposition.

This experiment asks a narrow question:

> Can consequential play leave a bounded semantic residue that preserves qualitative causal history after detailed simulation objects are compacted away?

## Scope

Base: current `master` `e9852845601606db97fc7a2931b69fd7d1af6d23`.

Exactly two added files:

- `src/js/gameplay/worldResidue.ts`;
- `experiments/storybook/src/Foundations/Strategy/WorldResidue.stories.ts`.

No production call sites, balance constants, Babylon/Cannon objects, terrain meshes, save/load formats, package/dependency metadata, or frozen #92 certification heads are changed.

## Pure residue model

`worldResidue.ts` separates detailed current records from compacted history.

Semantic event families:

- impact;
- wreck;
- fortress;
- geology;
- energy.

Representative event kinds include minor/heavy impacts, mothership hulks, fortress remnants, eruptions, geothermal depletion and extraction depletion.

### Aggregation

Ordinary events aggregate by region + causal family. Repeated events increase a saturating qualitative intensity rather than creating one permanent record per collision.

Landmarks currently include:

- mothership hulk;
- fortress remnant.

They retain stable event identity and receive higher retention priority than ordinary aggregate residue.

### Bounded compaction

`WorldResidueConfig.maxRecords` places a hard bound on detailed semantic records.

When the detailed budget is exceeded, the lowest-retention record is folded into a fixed-size `CompactedWorldResidue` summary that preserves:

- represented event count by family;
- strongest retained qualitative intensity by family.

The compacted layer intentionally loses exact event order, identity and location. This is the desired distinction between **historical meaning** and a perfect forensic log.

Retention preference considers:

- landmark status;
- current gameplay consequence;
- qualitative intensity;
- recency.

This is an experiment policy, not a final save/persistence algorithm.

## Storybook witness

Adds `Foundations/Strategy/World Residue`.

It presents the **same abstract arena after six different histories**:

1. pristine;
2. repeated heavy fire;
3. failed mothership crash;
4. geothermal exhaustion/eruption;
5. recent extraction;
6. abandoned fortress.

The layouts are intentionally geometrically equivalent. Only systemic residue differs, so reviewers can judge whether history is inferable without a log or explanatory cutscene.

The visual grammar uses redundant border/pattern/shape distinctions; color is not required for the semantic comparison.

### Compaction stress

A deterministic stress witness submits 180 ordinary events across many regions under a six-record cap, then adds a late mothership hulk.

The test verifies:

- detailed records remain `<= 6`;
- compacted history is non-empty;
- all 181 events remain represented between detailed + compacted state;
- the late major hulk remains explicit.

## Automated Storybook invariants

The story verifies:

- six histories render and have distinct semantic signatures;
- pristine carries no impact residue;
- repeated heavy fire crosses the contested-impact threshold;
- crash history preserves a mothership-hulk landmark;
- eruption produces strong geological history;
- extraction produces strong energy aftermath;
- abandoned fortress preserves a fortress-remnant landmark;
- compaction remains bounded while represented-event accounting is preserved.

## Design boundary

This model is **not** simulation authority.

It does not decide:

- crater geometry;
- collision/obstruction;
- exact visual meshes;
- hulk physics;
- terrain recovery;
- how long a residue should persist in campaign time;
- what save format should store it;
- whether a residue affects gameplay or is presentation-only.

Those remain with #79/#89/#76 and future adapters. This PR only supplies a compact semantic history seam that those systems could feed.

## Why this matters to the inversion

If later adapters preserve the same semantic causes, the player can encounter failed motherships, depleted targets and abandoned rooted defenses before fully understanding them, then reinterpret those traces after becoming a mothership themselves.

That strengthens the defender → raider reveal through recognition rather than exposition.

## Validation / promotion

Keep draft and outside all currently frozen certification cuts.

A later post-freeze campaign should run:

```sh
npm run lint
npm run build
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Also inspect `World Residue` at representative compact/desktop viewports and grayscale/reduced visual conditions where useful.

Before production use:

- calibrate event inputs from #79/#89/#76 rather than inventing duplicate physics/economy facts;
- test many-landmark overflow, event-time discontinuities and malformed/rehydrated state;
- decide which residues are gameplay-consequential versus presentation-only;
- measure long-horizon state/render budgets;
- add reload/reconstruction evidence;
- run #137's human comprehension/reinterpretation tests.

This is a semantic feasibility experiment, not a persistence-system commitment.

### DIFF
--- a/experiments/storybook/src/Foundations/Strategy/WorldResidue.stories.ts
+++ b/experiments/storybook/src/Foundations/Strategy/WorldResidue.stories.ts
@@ -1,0 +1,284 @@
+import type { Meta, StoryObj } from "@storybook/html-vite";
+import { expect } from "storybook/test";
+import {
+	appendWorldResidueEvent,
+	buildWorldResidueState,
+	createWorldResidueState,
+	summarizeWorldResidue,
+	type WorldResidueEvent,
+	type WorldResidueState,
+	type WorldResidueSummary
+} from "@defend/gameplay/worldResidue";
+import { createLabShell } from "../../labTheme";
+
+type HistoryFixture = {
+	id: string;
+	label: string;
+	description: string;
+	state: WorldResidueState;
+	summary: WorldResidueSummary;
+};
+
+function event(
+	id: string,
+	regionId: string,
+	kind: WorldResidueEvent["kind"],
+	occurredAtSeconds: number,
+	magnitude: number,
+	gameplayConsequence = false
+): WorldResidueEvent {
+	return { id, regionId, kind, occurredAtSeconds, magnitude, gameplayConsequence };
+}
+
+function repeatedHeavyFire(): WorldResidueEvent[] {
+	const events: WorldResidueEvent[] = [];
+	for (let index = 0; index < 10; index += 1) {
+		events.push(
+			event(
+				`heavy-${index}`,
+				index % 2 === 0 ? "east-approach" : "north-east-approach",
+				index % 3 === 0 ? "heavy-impact" : "minor-impact",
+				12 + index * 4,
+				index % 3 === 0 ? 0.5 : 0.24,
+				true
+			)
+		);
+	}
+	return events;
+}
+
+function fixture(
+	id: string,
+	label: string,
+	description: string,
+	events: WorldResidueEvent[]
+): HistoryFixture {
+	const state = buildWorldResidueState(events, { maxRecords: 12 });
+	return { id, label, description, state, summary: summarizeWorldResidue(state) };
+}
+
+function fixtures(): HistoryFixture[] {
+	return [
+		fixture("pristine", "Pristine", "No consequential history has accumulated yet.", []),
+		fixture(
+			"contested",
+			"Repeated heavy fire",
+			"The same approaches have absorbed repeated projectile impacts and now read as contested terrain.",
+			repeatedHeavyFire()
+		),
+		fixture(
+			"crash",
+			"Failed mothership",
+			"A large embodied commitment lost lift and remained as a recognizable landmark after impact.",
+			[
+				event("hulk-alpha", "west-edge", "mothership-hulk", 42, 0.96, true),
+				event("crash-impact", "west-edge", "heavy-impact", 42.2, 0.9, true),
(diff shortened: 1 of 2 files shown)


xtreemze (commented):
World-residue review: the bounded semantic-history direction is strong, but two corrections would make it a safer foundation for later narrative reinterpretation.

1. **Retention recency is relative to the appended event, not the actual newest known time.** When compaction runs, `lowestRetentionIndex(records, event.occurredAtSeconds)` uses the incoming event timestamp as `newestTime`. If events are replayed/reconstructed out of chronological order, an old event can make all later records appear age 0 and change which record is evicted. Use a monotonic/history-wide newest timestamp (e.g. max incoming time and retained last-occurrence times, or explicit state clock) when scoring retention. Add an unsorted/replayed-event fixture proving compaction result is stable with respect to ingestion order where semantic event set/timestamps are identical.

2. **Compacted state currently preserves historical texture, not causal history.** Family event counts + saturated family intensity cannot distinguish bounded but narratively different histories once detailed records are gone—for example fortress→mothership-wreck versus wreck→later-fortress, or repeated extraction around a geological collapse versus unrelated energy/geology events elsewhere. Since #137's inversion relies on later reinterpretation of prior causes, consider retaining a *small bounded causal motif layer* rather than a full log: a few landmark-linked/co-located family relationships or coarse transition signatures with strict caps. That would preserve "this wreck is part of what happened here" without retaining exact forensic history.

If the deliberate scope is only ambient historical texture, clarify that wording and keep causal reinterpretation dependent on explicit retained landmarks. If this model is meant to preserve qualitative causal history, family totals alone are too lossy after compaction.

Secondary robustness note: `finite()` still assumes runtime numeric types, so malformed rehydrated timestamps/magnitudes can propagate through some sanitizers as NaN. The PR already lists malformed/rehydrated state as pre-production work; use the #134/#141 runtime-number pattern when that hardening lands.

PR #144, opened 2026-09-04
Bootstrap one modern pnpm workspace for browser labs

xtreemze:
Refs #143
Parent direction: #66
Related: #31, #67, #149, #152, #155, #158

## Purpose

Establish one modern JavaScript workspace for the already-merged Storybook and hybrid-engine labs without touching the historical npm/Webpack application.

The two modern packages previously repeated Node/pnpm metadata but had no parent workspace or committed lockfile. This PR centralizes that ownership so the local executor can resolve one dependency graph and commit one lockfile before dependency-family upgrades begin.

## Scope

Current exact head: `e367d1abad6018ccd884adda4220de4f5d37b49a`.

Branch origin predates later `master` legacy-tooling deletions (#151/#156/#157/#160/#161). Those are disjoint from `experiments/**`; do not mutate this certification head merely to remove its behind count.

- add `experiments/package.json` as the modern workspace root;
- pin `packageManager` and `engines.pnpm` to pnpm 11.25.0;
- constrain Node to 24.20.x and add `.node-version` 24.20.0;
- add `experiments/pnpm-workspace.yaml` for exactly Storybook + hybrid-engine;
- centralize package-manager/runtime ownership out of child manifests;
- add hybrid `typecheck` for uniform aggregate verification;
- update both READMEs to install/run from the shared workspace;
- no direct dependency version changes.

## Fail-closed pnpm policy

Project settings live in `pnpm-workspace.yaml` under pnpm 11. The workspace explicitly configures:

```yaml
saveExact: true
sharedWorkspaceLockfile: true
engineStrict: true
minimumReleaseAge: 1440
minimumReleaseAgeStrict: true
minimumReleaseAgeIgnoreMissingTime: false
strictDepBuilds: true
verifyDepsBeforeRun: error
blockExoticSubdeps: true
trustLockfile: false
```

Consequences:

- direct dependencies remain exact-pinned;
- one shared lockfile is authoritative;
- wrong Node versions fail under `engineStrict` and wrong pnpm versions fail locally through exact `engines.pnpm`;
- newly published dependency versions must age at least one day and resolution fails rather than bypassing the gate;
- missing registry publish-time metadata fails closed;
- unreviewed dependency build/postinstall scripts fail the install and must be explicitly approved/denied;
- transitive exotic git/tarball sources remain blocked;
- `pnpm run` / `exec` will not silently auto-install stale dependencies;
- lockfile entries continue to receive supply-chain verification rather than being implicitly trusted.

[...]

Issue #143, opened 2026-09-04
Establish one reproducible modern pnpm workspace and lockfile


xtreemze:
Parent: #66
Related: #31, #67

## Purpose

Turn the already-merged modern experiments into one reproducible JavaScript workspace without modifying the historical npm/Webpack application.

The current state has two modern package manifests (`experiments/storybook` and `experiments/hybrid-engine`) that independently pin the same Node 24.20.x / pnpm 11.25.0 baseline, but there is no workspace root and no committed pnpm lockfile. This makes transitive dependency resolution and cross-experiment upgrade work less reproducible than the architecture in #66 intends.

## Ownership boundary

First slice should own only modern workspace/tooling metadata under `experiments/`:

- add an `experiments/package.json` workspace root with the canonical Node/pnpm baseline and non-destructive aggregate scripts;
- add `experiments/pnpm-workspace.yaml` covering the current modern packages;
- remove duplicated package-manager/runtime ownership from child manifests only if doing so keeps standalone intent clear;
- generate and commit one `experiments/pnpm-lock.yaml` from the real package manager during local certification.

Do **not** change in this slice:

- root historical `package.json`, npm/Webpack 4 dependency graph, or legacy build scripts;
- production gameplay/runtime code;
- generated PWA/Cordova artifacts;
- Babylon/physics/audio major architecture;
- Rust workspace membership or Cargo profiles/lockfile (track separately because the hybrid crate currently carries release-profile semantics that need deliberate workspace treatment).

## Dependency policy

- Keep direct modern dependencies exactly pinned during certification.
- Use one shared lockfile for transitive resolution.
- Upgrade deliberately by family, with Babylon packages kept mutually compatible.
- Do not merge legacy-root Dependabot updates merely to make the historical graph look current; #31 remains authoritative for coherent migration.

As of 2026-09-04, most modern tool pins remain current; Babylon core has moved from 9.23.0 to 9.25.0. Treat that as a follow-up family upgrade after the shared lockfile exists rather than combining workspace establishment with a just-published renderer update.

## Local certification / completion gate

The local executor should, from `experiments/`, use Node 24.20.x and pnpm 11.25.0 to:

1. run `pnpm install` and inspect the resolved graph;
2. commit the generated shared `pnpm-lock.yaml` only if resolution is clean;
3. run aggregate type/check/test/build commands;
4. ensure Storybook typecheck/build/browser tests remain healthy;
[...]

Issue #49, opened 2026-09-03
Quarantine or remove destructive legacy release helpers


xtreemze:
Parent: #31
Related: #35, #48

## Findings

The remaining repository helper/release surface contains commands that are inappropriate for automated executors and, in places, appears broken:

- `package.json` script `posttoTest` performs `git add -A`, forces a version bump, invokes `git push` and `git push --tags`, and references `npm run commits` even though no `commits` script exists;
- `node/clean.js` deletes the entire `node_modules` directory;
- `node/rmDir.js` deletes `dist/` and is invoked by `preexportp`;
- `node/archive.js` writes versioned ZIPs into `release/`/`dist/release/` and requires `archiver`, but repository search shows no declared `archiver` dependency in `package.json`;
- the historical Travis deployment that invoked this export path has now been retired in #48.

These scripts may have been useful in the original manual release workflow, but they should not remain ambiguous commands that an AI executor might treat as ordinary validation/build operations.

## Proposed policy

After #35 settles the safe default `lint`/`build` semantics:

1. remove or explicitly rename destructive/manual-only scripts with a `legacy:` or `release:` namespace;
2. eliminate scripts that stage/commit/push Git state from `package.json`; release publication should never implicitly mutate the caller's checkout;
3. remove dead helper files once no maintained script references them;
4. rebuild any retained packaging operation as a deterministic function of a clean build output, with declared dependencies and no Git side effects;
5. make generated output locations explicit and ignored where appropriate;
6. document which commands are safe for automated validation versus intentional publication/destructive maintenance.

## Constraints

- do not overlap #35 while it owns `package.json` command semantics;
- do not introduce a replacement deployment workflow until the local historical-build baseline is known;
- do not preserve a broken helper merely because it is old;
- do not add CI/publishing credentials or automatic pushes as part of cleanup.

This is repository-process modernization only; no gameplay/runtime behavior is intended to change.

xtreemze:
## 2026-09-04 cleanup status

The unsafe/dead portion of this issue is now substantially resolved on `master`:

- #151 removed `posttoTest` and all implicit Git stage/version/push behavior;
- #156 removed dead `node/clean.js` (`node_modules` recursive deletion);
- #157 removed unreferenced release-copy helpers;
- #160 removed the broken `archive` path plus unpinned `npx pwabuilder` generators;
- #161 removed the broken ESLint config-check command;
[...]

PR #151, opened 2026-09-04
Remove destructive legacy Git-push release script

xtreemze:
Refs #49
Parent: #31
Related: #35, #48, #143, #144

## Purpose

Remove the remaining root package command that can unexpectedly mutate and publish repository state.

`posttoTest` currently performs all of the following from an ordinary npm script:

- changes npm Git-tag configuration;
- `git add -A` across the entire checkout;
- force-bumps the package version;
- invokes `npm run commits`, which is not declared;
- pushes the current branch;
- pushes Git tags.

That behavior is incompatible with the repository's multi-executor safety model and is already effectively broken because the referenced `commits` script does not exist.

## Change

Delete only the `posttoTest` script entry from the historical root `package.json`.

## Deliberately unchanged

- no dependency versions or lockfiles;
- no build/lint/dev commands;
- no archive/PWA/export helpers yet;
- no gameplay, rendering, physics, audio, or generated output;
- no modern `experiments/` workspace metadata.

The remaining legacy packaging helpers stay tracked under #49 for deliberate quarantine/removal after their usage is characterized. This PR removes only the command that can stage/version/push caller state.

## Validation

Static scope is one deleted JSON property. `package.json` remains valid JSON, and no maintained script references `posttoTest`.

No runtime/browser certification is required because this removes an unused destructive publication command and does not alter any executable validation/build path.

## Promotion

Safe to merge after exact-head static review. Do not broaden this PR into historical dependency upgrades.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
         "exportp": "webpack --mode production && npm run favicon && npm run manifest",
         "build": "webpack --mode production",
         "build:watch": "webpack --mode production --watch",
-        "posttoTest": "npm config set sign-git-tag true && git add -A && npm version patch -s --force && npm run commits & npm run queryVersion && git push && git push --tags",
         "favicon": "real-favicon generate faviconDescription.json faviconData.json ./dist/ && real-favicon inject faviconData.json ./dist/ ./dist/index.html",
         "faviconUpdate": "real-favicon check-for-update --fail-on-update faviconData.json",
         "screenshot": "node ./node/puppeteer.js"


xtreemze (commented):
Exact-head static review PASS. Diff is exactly one deleted `package.json` script property. The removed command stages the entire checkout, force-bumps version state, references an undeclared `commits` script, and pushes commits/tags. No dependency, lockfile, build, runtime, or gameplay surface changes. Safe to merge without runtime certification.

Issue #154, opened 2026-09-04
Move automated dependency updates to maintained modern workspaces


xtreemze:
Parent: #66
Depends on: #143 / #144
Related: #31, #145, #149, #152

## Purpose

Once the modern workspace has a real shared lockfile, configure automated dependency discovery around maintained source rather than the historical root and generated PWA trees.

The stale root Dependabot PRs #69/#70/#71 are now closed and their security obligations are preserved in #152. There is currently no `.github/dependabot.yml` on `master`.

### Modern pnpm workspace

Target `/experiments` only until the final `/apps` + `/packages` layout exists.

Use deliberate grouping so related packages move together:

- **Babylon family:** `@babylonjs/core`, `@babylonjs/inspector`, `@babylonjs/accessibility`;
- **Storybook family:** `storybook`, `@storybook/*`;
- **Vite/Vitest browser tooling:** `vite`, `vitest`, `@vitest/*`, Playwright packages;
- **Type/lint tooling:** `typescript`, `@types/*`, `@biomejs/biome`.

Do not auto-merge. Exact pins and the shared lockfile remain authoritative. Let pnpm's package-maturation protections apply rather than immediately taking just-published releases.

### Rust

Until #145 consolidates the Cargo workspace, dependency discovery has two distinct Rust roots:

- `/` for `crates/defend-core`;
- `/experiments/hybrid-engine/rust` for the standalone hybrid runtime.

Prefer to delay permanent Dependabot Cargo configuration until workspace/profile ownership and one Cargo.lock are settled. If enabled earlier, keep the two roots explicit and never assume one update covers both.

### Explicit exclusions

Do not create ordinary dependency-update PRs for:

- historical root npm/Yarn/Webpack/Babylon 3 graph;
- `release/pwa/**` generated Cordova trees;
- archived/generated output.

Those surfaces are governed by #152 and eventual retirement/regeneration, not ongoing modernization.

## Update cadence

Weekly is sufficient for ordinary dependency discovery. Security alerts remain separate. Avoid daily PR churn while experimental APIs and browser/runtime certification are expensive.

## Completion gate

After #144 merges:

1. add the minimal Dependabot config;
2. validate the config syntax and target directories;
3. confirm the first generated PRs target only maintained modern manifests;
4. confirm grouping does not combine renderer, physics authority, and unrelated tooling changes;
5. update the config when `experiments/*` migrates into the final `apps/packages` structure.

[...]

Issue #152, opened 2026-09-04
Contain legacy dependency security risk while the historical root remains executable


xtreemze:
Parent: #31
Related: #49, #69, #70, #71, #143, #144

## Why this exists

The historical root still matters as a behavioral-parity reference during modernization, but it is not the dependency graph we intend to carry forward.

The remaining Dependabot PRs mix three different concerns:

- #69 updates root `browserslist` and a broad set of transitive browser-data packages inside the historical Yarn graph;
- #71 changes only the historical `yarn.lock` to move transitive `js-yaml` 3.13.1 -> 3.15.2, carrying several security/backported complexity/prototype-pollution fixes;
- #70 upgrades generated Cordova-Android material under `release/pwa/...` from 7.1.4 to 15.1.0, a many-major generated-artifact migration that should not be merged as if it were ordinary application source.

Per #31 these PRs are security/obsolescence evidence, not a required merge sequence.

## Historical dependency ambiguity

The root has a committed historical `yarn.lock` but no committed npm lockfile, while local certification currently records the actually resolved historical install rather than treating a newly generated npm lock as source. That means a lockfile-only Dependabot patch such as #71 may not protect the environment actually used for certification.

We should not create false confidence by declaring the legacy graph secure because one stale lockfile was patched.

## Containment policy until retirement

1. Treat the root dependency graph as **legacy/certification-only**, not a source for new application packages.
2. Do not add new production features that depend on the historical Webpack/Babylon/Cannon/Tone graph.
3. During the next local historical install, record the package manager, exact resolved tree, audit/security output, and whether the committed `yarn.lock` is actually authoritative for that lane.
4. If a high-severity issue is reachable during normal build/certification, make the smallest isolated compatibility/security patch required to keep the legacy baseline usable; do not opportunistically modernize the whole graph.
5. Keep untrusted external YAML/data away from historical build tooling where practical; #71 shows the old graph contains parser versions with known resource-exhaustion/prototype-pollution history.
6. Treat generated Cordova/PWA trees as artifacts pending the dedicated PWA cleanup from #31, not as a dependency workspace to upgrade in place.
7. Retire the historical graph once the modern browser route reaches the behavioral parity gate; at that point remove the legacy lockfile/security surface rather than continuing backports indefinitely.

## Local evidence requested

On the next legacy-root certification target:

- report Node + npm/yarn versions;
- identify the install command actually capable of reconstructing the historical application;
[...]

Issue #149, opened 2026-09-04
Align the modern Babylon 9.x package family after workspace lockfile lands


xtreemze:
Parent: #66
Depends on: #143 / #144
Related: #31, #67, #145

## Trigger

The hybrid lab currently pins Babylon ESM packages at 9.23.0. On 2026-09-04, `@babylonjs/core` and the Inspector family published 9.25.0. The renderer patch should be taken as a coherent Babylon-family upgrade **after** #144 establishes the shared modern pnpm lockfile, not mixed into workspace bootstrapping.

Upstream package evidence checked today:

- `@babylonjs/core` 9.25.0: https://www.npmjs.com/package/@babylonjs/core
- Inspector 9.25.0 release family: https://www.npmjs.com/package/babylonjs-inspector
- `@babylonjs/havok` remains 1.3.14: https://www.npmjs.com/package/@babylonjs/havok

## Scope

After #144 is locally certified and its shared `experiments/pnpm-lock.yaml` exists:

1. resolve the current stable versions with the real package manager (`pnpm view`) for `@babylonjs/core`, `@babylonjs/inspector`, `@babylonjs/accessibility`, and `@babylonjs/havok`;
2. keep the core/Inspector/accessibility packages aligned to a compatible Babylon 9.x release family rather than bumping one package independently;
3. leave Havok/Rapier physics ownership unchanged unless a compatibility constraint requires a patch-level adjustment;
4. update the shared lockfile from `experiments/` only;
5. do not change gameplay, simulation authority, physics API generation, or production legacy Babylon 3 code in this PR.

## Certification

- `pnpm install --frozen-lockfile` after the lock update;
- workspace typecheck/check/test/build;
- hybrid browser fixture renders all bodies without console errors;
- headless Inspector path (`?inspect=1`) still connects, queries scene state, and tears down cleanly;
- accessibility package imports/fixtures still resolve;
- compare raw/compressed JS payload against the 9.23.0 baseline;
- capture WebGL2 baseline behavior and WebGPU capability path if locally available;
- record any new deprecations or changed tree-shaking/import behavior;
- keep Physics V1/V2/Havok/Rapier authority decisions outside this patch upgrade.

## Promotion rule

Use a small draft PR based on the post-#144 workspace head and attach local browser/build evidence before promotion. If 9.25.0 proves incompatible, record the exact blocker here rather than widening the PR into an engine migration.

xtreemze:
## Release-maturation gate

[...]

Issue #145, opened 2026-09-04
Pin Rust 1.98.1 and separate Cargo workspace consolidation


xtreemze:
Parent: #66
Related: #31, #67, #143, #144

## Immediate toolchain correction

Rust 1.98.1 was released on 2026-09-03 specifically to fix a 1.98.0 compiler miscompilation in trait-object vtable generation that could place a null pointer where a function pointer should be, leading to undefined behavior. The hybrid lab currently records Rust 1.98.0 / edition 2024 and its crate uses `rust-version = "1.98"`.

Before doing wider Rust workspace/dependency work, pin the repository's modern Rust baseline to **1.98.1** so local certification cannot accidentally use the affected compiler.

Recommended narrow PR:

- add a repository `rust-toolchain.toml` pinned to 1.98.1;
- include `rustfmt` and `clippy` components;
- include `wasm32-unknown-unknown`, required by the hybrid WASM path;
- raise the private Rust crates' project MSRV to 1.98.1 where appropriate;
- update modern documentation from 1.98.0 to 1.98.1;
- do not change Bevy/Rapier/wasm-bindgen versions or workspace membership in that PR.

## Separate Cargo workspace problem

Current root `Cargo.toml` has only:

```toml
[workspace]
members = ["crates/defend-core"]
resolver = "2"
```

The hybrid runtime at `experiments/hybrid-engine/rust` remains standalone and currently owns a local `[profile.release]` with size/LTO/panic settings. There is also no committed root `Cargo.lock`.

Simply adding that crate as a root workspace member is not a metadata-only change: Cargo profile ownership moves to the workspace root, so we must preserve the hybrid WASM release semantics deliberately rather than silently losing or broadening them.

## Cargo consolidation questions

After the 1.98.1 pin is certified:

1. decide whether `experiments/hybrid-engine/rust` should become a temporary root workspace member before its eventual `crates/defend-runtime` relocation;
2. preserve the hybrid runtime's release-size/LTO/panic behavior explicitly at the workspace root or via a dedicated build profile;
3. decide which package metadata belongs under `[workspace.package]`;
4. rationalize `wasm-bindgen` ownership (`defend-core` currently uses an optional `0.2` range; hybrid pins 0.2.127) without changing runtime behavior accidentally;
5. generate and commit one real `Cargo.lock` from Cargo 1.98.1;
6. run `cargo metadata`, `cargo tree`, fmt/clippy/test and WASM builds and confirm no Bevy renderer/window stack appears;
[...]

PR #155, opened 2026-09-04
Point Dependabot at the maintained modern workspace

xtreemze:
Refs #154
Parent: #144 / #143
Related: #31, #66, #149, #152, #158

## Purpose

Move automated JavaScript dependency discovery onto the maintained modern pnpm workspace and stop using the historical root/generated PWA trees as the modernization surface.

This PR is intentionally stacked on #144 because `/experiments` becomes a real workspace/lockfile root there. Do not retarget/merge it to `master` before #144 lands and its lockfile is locally certified.

Current exact head: `b6cdabd116aab21a3a5f29a297d7eac5ee04a8f1`.

It has been reassembled on #144 exact head `e367d1abad6018ccd884adda4220de4f5d37b49a` with a normal merge commit preserving the original policy commit as ancestry. Comparison against the current parent is exactly one added file: `.github/dependabot.yml`.

## Configuration

Adds one npm ecosystem rooted at `/experiments`, checked weekly.

Groups:

- `babylon`: all `@babylonjs/*` packages move together;
- `storybook`: `storybook` + `@storybook/*` move together;
- `vite-vitest-browser-tooling`: Vite, Vitest, `@vitest/*`, Playwright patch/minor updates grouped;
- `type-and-lint-tooling`: TypeScript, `@types/*`, Biome patch/minor updates grouped.

Tooling major versions are intentionally not folded into those broad groups; they should surface as separate migration PRs. Babylon/Storybook remain family-grouped because package-family alignment is itself part of compatibility.

`versioning-strategy: increase` is used because the modern workspace is an application/test workspace with exact direct pins rather than a published compatibility-range library.

## Deliberately absent

- no root `/` npm update entry;
- no `release/pwa/**` Cordova update entry;
- no Cargo automation yet;
- no auto-merge;
- no dependency changes in this PR;
- no pnpm catalog migration while Dependabot catalog/lockfile correctness remains under observation.

Rust automation waits for #145 to decide Cargo workspace/profile/lockfile ownership. Legacy dependency risk is tracked by #152 rather than continuously upgrading the 2019 graph. pnpm-major migration is #158.

## Validation / promotion

Current GitHub Dependabot configuration docs support `version: 2`, `directory`, weekly schedules, `groups.patterns`, and group `update-types`.

After #144 merges:

1. retarget/reassemble this one-file delta onto current `master` without changing config semantics;
2. confirm GitHub accepts the configuration in the Dependency graph / Dependabot surface;
3. trigger or wait for the first version-update cycle;
4. verify generated PRs touch only `/experiments` manifests/shared lockfile;
5. verify Babylon and Storybook family grouping behaves as intended;
6. verify tooling majors remain separate from ordinary patch/minor churn;
7. verify update PRs preserve #144's strict pnpm lockfile/supply-chain policy;
8. keep draft until those repository-level checks are recorded.

This config supplies dependency intelligence; it does not replace pnpm's shared lockfile, release-maturation/build-script policy, or local browser/build certification.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,38 @@
+version: 2
+
+updates:
+  - package-ecosystem: npm
+    directory: /experiments
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 8
+    versioning-strategy: increase
+    groups:
+      babylon:
+        applies-to: version-updates
+        patterns:
+          - "@babylonjs/*"
+      storybook:
+        applies-to: version-updates
+        patterns:
+          - "storybook"
+          - "@storybook/*"
+      vite-vitest-browser-tooling:
+        applies-to: version-updates
+        patterns:
+          - "vite"
+          - "vitest"
+          - "@vitest/*"
+          - "playwright"
+        update-types:
+          - patch
+          - minor
+      type-and-lint-tooling:
+        applies-to: version-updates
+        patterns:
+          - "typescript"
+          - "@types/*"
+          - "@biomejs/biome"
+        update-types:
+          - patch
+          - minor
--- a/experiments/package.json
+++ b/experiments/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "packageManager": "pnpm@11.25.0",
   "engines": {
-    "node": "24.20.x"
+    "node": "24.20.x",
+    "pnpm": "11.25.0"
   },
   "scripts": {
     "typecheck": "pnpm -r --workspace-concurrency=1 run typecheck",
--- a/experiments/pnpm-workspace.yaml
+++ b/experiments/pnpm-workspace.yaml
@@ -5,3 +5,6 @@ packages:
 saveExact: true
 sharedWorkspaceLockfile: true
 engineStrict: true
+minimumReleaseAge: 1440
+minimumReleaseAgeStrict: true
+minimumReleaseAgeIgnoreMissingTime: false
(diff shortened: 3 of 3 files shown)


xtreemze (commented):
Static dependency-automation review found one external compatibility blocker that should gate promotion.

GitHub's current Dependabot supported-ecosystems/options documentation lists pnpm support under `package-ecosystem: npm` for **pnpm v7-v10**, while parent #144 intentionally pins the maintained workspace to **pnpm 11.25.0**. The grouping syntax here (`applies-to`, `patterns`, `update-types`) is supported, and `/experiments` is structurally the right ownership boundary, but we should not merge a config whose lockfile/package-manager generation is outside GitHub's documented supported range and then treat missing/broken update PRs as a repository defect.

Recommendation:

1. keep #155 draft/blocked while #144 remains on pnpm 11;
2. do **not** downgrade the modern workspace solely to make Dependabot work;
3. after #144 is locally certified, re-check GitHub's support table before promotion — pnpm 11 support may arrive independently;
4. if it still lists only v7-v10, track the config here/#154 but leave it unmerged, or evaluate a separate dependency-discovery mechanism without changing package-manager ownership;
5. once pnpm 11 is officially supported, validate one real Dependabot run against the committed shared lockfile before marking this ready.

Minor note for that future activation: the current workspace uses `playwright`, so the browser-tooling pattern is correct today; if `@playwright/test` is later introduced, add `@playwright/*` to keep the family grouped.

No issue with the YAML grouping semantics themselves was found.

xtreemze:
Fulfilled on current `master`: `.github/dependabot.yml` now targets `/experiments` with the intended Babylon, Storybook, Vite/Vitest/Playwright, and type/lint groups, and the resulting Dependabot PRs have already demonstrated that GitHub accepts and executes this configuration. Closing the stale stacked configuration PR as provenance.

Issue #31, opened 2026-09-03
Modernize the toolchain and dependencies as a coherent migration, not stale Dependabot merges


xtreemze:
Parent: #29
Depends on findings from #30, but investigation and migration design can proceed in parallel.

## Problem

The repository is pinned to a 2019-era stack (BabylonJS 3.x, Cannon 0.6, TypeScript 3.2, Webpack 4 and related tooling) and has a backlog of years-old Dependabot PRs. Merging those PRs individually would preserve an obsolete dependency graph while increasing migration risk.

## Goal

Move Defend onto a current, maintainable web-game toolchain in deliberate stages while preserving the behavior contract in #29/#30.

## Principles

- Prefer replacing obsolete build infrastructure over accumulating compatibility shims.
- Keep gameplay changes out of the toolchain PR unless required to restore equivalent behavior.
- Preserve direct browser/PWA delivery.
- Prefer modular imports/tree shaking and a small production bundle.
- Keep source maps/debuggability useful for AI executors and local browser inspection.
- Establish reproducible package-manager and runtime versions.
- Add modern lint/type/test scripts that are non-destructive by default.
- Treat the stale Dependabot PRs as historical security signals, not as a sequence that must be merged.

## Investigation questions

- Vite vs another minimal modern bundler for BabylonJS/web-worker/asset requirements.
- Current BabylonJS migration path and whether to adopt modular `@babylonjs/*` packages immediately or in a later PR.
- Physics migration path: legacy Cannon plugin compatibility vs Cannon-es/Havok/another Babylon-supported engine. Preserve physical gameplay before changing the engine.
- PWA/service-worker modernization and whether generated legacy PWA artifacts under `release/` should remain tracked.
- Whether procedural audio should remain on the existing vendored/Tone-era implementation temporarily or be isolated behind a small adapter first.

## Suggested PR sequence

1. Reproducible package/runtime metadata and non-invasive scripts.
2. Modern build/dev server while keeping the current game code behaviorally equivalent.
3. TypeScript/lint modernization and removal of obsolete configuration.
4. Babylon modularization/upgrade behind #30 certification.
5. Physics engine upgrade as its own measured PR.
6. PWA/service-worker cleanup.
7. Close/supersede stale dependency PRs once their concerns are covered.

## Local Codex role

The local executor should run installs/builds, measure bundle/runtime behavior, verify browser compatibility and performance, and report exact migration blockers back to this issue and each PR.

xtreemze:
[...]

PR #156, opened 2026-09-04
Remove dead destructive node_modules cleanup helper

xtreemze:
Refs #49
Parent: #31
Related: #151, #152

## Purpose

Remove a dead helper whose only behavior is recursively deleting the repository's `node_modules` directory.

`node/clean.js` is not exposed through the root `package.json`, and repository search finds no caller. Keeping an unreferenced destructive helper adds risk for automated executors without preserving a maintained workflow.

## Change

Delete exactly `node/clean.js`.

## Deliberately unchanged

- `preexportp` / `node/rmDir.js` remain because they are still referenced by the legacy export path;
- `archive` / `node/archive.js` remain because the package script still references them;
- no dependencies, lockfiles, builds, generated output, gameplay, or runtime code change.

## Validation

Static-only change. Search of current `master` finds no reference to `clean.js`, and the deleted file only invoked `fs-extra.remove("./node_modules")`.

No local runtime certification is required for deletion of an unreachable destructive helper.

### DIFF
--- a/node/clean.js
+++ b/node/clean.js
@@ -1,5 +1,0 @@
-const fsExtra = require("fs-extra");
-
-fsExtra.remove("./node_modules", error => {
-  Error(error);
-});


xtreemze (commented):
Exact-head static review PASS. The diff deletes only an unreferenced helper whose sole action is recursive removal of `./node_modules`. No package script, dependency, build, runtime, or generated-output path changes. Safe to merge without runtime certification.

PR #157, opened 2026-09-04
Remove unreachable legacy release copy helpers

xtreemze:
Refs #49
Parent: #31
Related: #151, #156, #152

## Purpose

Remove two unreferenced legacy Node helpers that copy obsolete release artifacts but are not reachable from current package scripts or repository source.

Deleted:

- `node/cpDirProd.js` — copied `src/.htaccess` into `dist/.htaccess`;
- `node/cpReport.js` — copied `release/lighthouse.html` into `dist/release/lighthouse.html`.

Repository search finds no callers for either filename. Both also declare an unused package-version variable, consistent with abandoned historical release automation.

## Deliberately unchanged

- `node/cpDir.js` remains because `preexportp` still calls it;
- `node/rmDir.js` remains because `preexportp` still calls it;
- `node/archive.js` remains because the `archive` package script still calls it;
- no package scripts, dependencies, lockfiles, builds, generated output, or runtime/gameplay code change.

## Validation

Static-only deletion of two unreachable helpers. No local runtime certification required.


xtreemze (commented):
Exact-head static review PASS. Diff contains only deletion of the two repository-unreferenced copy helpers documented in the PR. Active `cpDir.js`, `rmDir.js`, archive/package scripts, dependencies, lockfiles, and runtime paths are untouched. Safe to merge without local runtime certification.

PR #147, opened 2026-09-04
Pin Rust 1.98.1 after compiler miscompilation fix

xtreemze:
Refs #145
Parent direction: #66
Related: #67, #143, #144

## Why this is urgent

Rust 1.98.1 was released on 2026-09-03 to fix a 1.98.0 compiler miscompilation in trait-object vtable generation. The Rust release team states that 1.98.0 could emit a null pointer where a function pointer should be, leading to undefined behavior.

Upstream: https://blog.rust-lang.org/2026/09/03/Rust-1.98.1/

Defend's modern hybrid lab was established against Rust 1.98.0. This PR prevents local/runtime certification from accidentally using the affected compiler while keeping the change independent from broader Cargo workspace/dependency work.

## Scope

- add root `rust-toolchain.toml` pinned to `1.98.1`;
- use rustup `minimal` profile;
- ensure `clippy` and `rustfmt` components are available;
- ensure the `wasm32-unknown-unknown` target used by the hybrid WASM path is available;
- set `defend-core`'s private project MSRV to `1.98.1`;
- raise the hybrid runtime's existing `rust-version` from `1.98` to `1.98.1`.

## Deliberately unchanged

- no Rust source code;
- no workspace membership changes;
- no `Cargo.lock` synthesis;
- no Bevy, wasm-bindgen, Rapier, or other crate version changes;
- no release-profile changes;
- no JS package/workspace files (that work is isolated in #144).

The hybrid README currently mentions 1.98.0. I am intentionally not touching that file here because #144 already owns it; whichever lane lands second should reconcile the documentation without creating a cross-PR conflict.

## Local certification required

Keep draft until local evidence is attached to this exact head:

1. `rustup show active-toolchain` resolves 1.98.1 from the repository toolchain file;
2. `rustc -Vv` and `cargo -V` confirm the expected compiler/toolchain;
3. `rustup component list --installed` includes rustfmt and clippy;
4. `rustup target list --installed` includes `wasm32-unknown-unknown`;
5. root `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test` succeed or blockers are documented;
6. hybrid-runtime fmt/clippy/test succeed;
7. the existing wasm-pack build succeeds;
8. compare the resulting hybrid WASM/browser fixture against the current baseline and confirm no regression attributable to the compiler point release;
9. record exact commands, exit codes, head SHA, and final working-tree status.

## Follow-up

[...]

PR #159, opened 2026-09-04
Remove stale nested Rust 1.98.0 toolchain override

xtreemze:
Refs #145, #92
Parent PR: #147

## Purpose

Fix a safety hole in #147 without modifying its executor-owned branch in place.

#147 adds a repository-root `rust-toolchain.toml` pinned to Rust 1.98.1, but the hybrid runtime already contains a nearer:

`experiments/hybrid-engine/rust/rust-toolchain.toml`

still pinned to **1.98.0**.

Rustup resolves checked-in toolchain files by proximity while walking from the current directory toward the root. Therefore commands executed inside the hybrid crate would continue selecting the affected 1.98.0 compiler and bypass #147's root safety pin.

## Change

Delete only the stale nested hybrid `rust-toolchain.toml`.

The parent root toolchain then becomes authoritative for both:

- `crates/defend-core`;
- `experiments/hybrid-engine/rust`.

The root file already supplies the hybrid's required `wasm32-unknown-unknown` target plus `clippy` and `rustfmt`, so no capability is lost.

## Why deletion rather than updating both files

The modernization direction is one reproducible Rust baseline for the repository. Keeping two identical pins would recreate drift risk on the next point release and complicate the Cargo-workspace consolidation tracked in #145.

## Scope

Relative to #147 exact head `91543cf8a340f95caaa4e01c46c7bf00d5b3f7c7`:

- exactly one file deleted;
- no Rust source changes;
- no Cargo dependency/workspace/profile changes;
- no JS/package changes;
- no frozen #95/#97/#105 source changes.

## Local certification

Prefer certifying #147 with this child applied, or fold the deletion into #147 before its exact-head local gate.

From repository root **and** from `experiments/hybrid-engine/rust`, record:

```sh
rustup show active-toolchain
rustc -Vv
cargo -V
```

Both locations must resolve Rust 1.98.1. Then run #147's root/hybrid fmt/clippy/test/WASM checks.

If this child is not applied, hybrid-runtime certification performed from its crate directory must not be accepted as evidence that #147 successfully eliminated Rust 1.98.0.

### DIFF
--- a/experiments/hybrid-engine/rust/rust-toolchain.toml
+++ b/experiments/hybrid-engine/rust/rust-toolchain.toml
@@ -1,5 +1,0 @@
-[toolchain]
-channel = "1.98.0"
-profile = "minimal"
-targets = ["wasm32-unknown-unknown"]
-components = ["clippy", "rustfmt"]


xtreemze:
## Certified via #163 — PASS

This one-file deletion was exercised through **#163** head `c1318a946a841faeb90e7ab6500fb8a3ee20b276`, which folds in this PR's cut (`0e7f3399aed25ad97fdb3a8e70a90dcd46d71fd8`) alongside #147. Per §7 this certifies the tested tree, not this PR's own head SHA.

The check this PR exists for — verifying the active toolchain from **both** the repository root and `experiments/hybrid-engine/rust`, per the "Correction to Rust safety preflight" note in #92 — passes:

| Location | `rustup show active-toolchain` |
|---|---|
| repository root | `1.98.1-aarch64-apple-darwin (overridden by <root>/rust-toolchain.toml)` |
| `experiments/hybrid-engine/rust` | `1.98.1-aarch64-apple-darwin (overridden by <root>/rust-toolchain.toml)` |

`ls experiments/hybrid-engine/rust/rust-toolchain.toml` → `No such file or directory`. The stale nested 1.98.0 override is gone, and rustup correctly falls through to the root pin from inside the crate directory rather than bypassing it. `rustc 1.98.1 (48a229cea 2026-09-01)`, `cargo 1.98.1 (797e8a9bc 2026-08-05)` from both locations.

No artifact in this campaign was built with 1.98.0.

Note for whoever runs the hybrid crate next: the toolchain resolves correctly, but no cargo command actually runs in that directory until #166 lands — see #165. That is unrelated to this deletion.

Full campaign evidence: https://github.com/xtreemze/defend/pull/163#issuecomment-5563899938

xtreemze:
Superseded on current `master`: the stale `experiments/hybrid-engine/rust/rust-toolchain.toml` override is gone and the repository-root Rust 1.98.1 toolchain is authoritative for the hybrid crate. Closing this corrective child as provenance rather than preserving a stale stacked branch.

PR #160, opened 2026-09-04
Quarantine obsolete legacy release generators

xtreemze:
Refs #49
Parent: #31
Related: #48, #151, #152, #157

## Purpose

Remove three non-certification release commands that are unsafe or broken in the historical root while preserving the build/export baseline.

### Removed PWA generators

The root `pwa` and `pwaWeb` scripts invoke `npx pwabuilder` without a pinned package version. The checked-in PWA artifacts report generation by PWA Builder `2.0.3-rc.0`; invoking an unpinned current CLI against that historical output is not a reproducible certification path and creates arbitrary registry-code execution from a legacy package script.

PWA regeneration remains a deliberate modernization step under #31/#152 after the modern browser output shape is stable.

### Removed archive generator

The `archive` script calls `node/archive.js`. That helper:

- requires `archiver`, which is not declared in the root dependency graph;
- writes versioned zip artifacts under `release/`;
- removes `dist/release/` during its finish path.

The command is therefore both broken from the committed manifest and capable of mutating release output. The script and unreachable helper are removed together.

## Deliberately preserved

- `build` / `build:watch` historical Webpack paths;
- `dev` / `soundTest`;
- `preexportp` / `exportp`, including `rmDir.js` + `cpDir.js`, because they remain part of the historical export baseline pending characterization;
- manifest/favicon/screenshot helpers;
- all dependencies and lockfiles;
- all checked-in PWA/generated artifacts (cleanup/regeneration is separate).

## Validation

Static scope only:

- remove `archive`, `pwa`, and `pwaWeb` script entries;
- delete only `node/archive.js`;
- no maintained script references `archive.js` after the change;
- no dependency/build/gameplay/runtime source change.

Safe to merge after exact-head static review. Any future PWA packaging should be rebuilt as a pinned modern workspace rather than restoring unpinned `npx` execution in the historical root.


xtreemze (commented):
Exact-head static review PASS. Diff removes only three non-certification release scripts and the broken archive helper they exposed. Historical build/dev/export scripts and dependencies are unchanged. The removed PWA commands were unpinned `npx` execution; the removed archive helper required undeclared `archiver` and mutated release output. Safe to merge without runtime certification.

PR #164, opened 2026-09-04
Refresh local certification guidance after tooling cleanup

xtreemze:
Refs #92
Related: #35, #49, #144, #147, #151, #156, #157, #160, #161

## Purpose

Bring the stable local-certification procedure into line with the repository's current tooling state.

The document still said #35 had not merged and warned that ordinary root lint/build/sound commands were unsafe. It also assumed Storybook would always resolve dependencies from its child directory, which is no longer sufficient once #92 targets a shared modern workspace candidate.

## Changes

- record that #35 has merged and describe the current safe semantics of root `lint`, `build`, `build:watch`, and `soundTest`;
- record the later removal of destructive/broken release helpers;
- warn that `exportp` remains intentionally mutating and should only run for explicit export characterization;
- preserve the historical lockfile/package-manager ambiguity as evidence to record, not silently rewrite;
- make the browser-lab procedure aware of package-local vs shared-workspace targets from #92;
- make Rust guidance respect a pinned repository toolchain when present;
- add supply-chain/cache inspection to the generic Rust lane when #92 requests it;
- replace the obsolete priority for future #35 certification with dependency/workspace/toolchain targets that unblock downstream work;
- keep #92 as the source of exact live ordering/SHAs.

## Scope

Documentation only. No scripts, dependencies, lockfiles, runtime, gameplay, or workspace state change.

## Validation

Static review against current `master` and the live #92 packet. Safe to merge without runtime certification.

### DIFF
--- a/docs/LOCAL_CERTIFICATION.md
+++ b/docs/LOCAL_CERTIFICATION.md
@@ -81,23 +81,31 @@ Capture at least:
 
 Prefer one coherent recorded browser session over repeated restarts for every issue.
 
-#### Legacy root script warning
+#### Historical root command status
 
-Until #35 is certified and merged, the root package still has historical unsafe command semantics:
+#35 has merged. The ordinary root validation commands now have safe semantics for certification:
 
-- `npm run lint` contains `--fix` and mutates source;
-- `npm run build` contains `--watch` and does not terminate normally;
-- `npm run soundTest` contains the invalid `--hots` flag.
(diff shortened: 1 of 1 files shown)


xtreemze (commented):
Exact-head documentation review PASS. The diff removes obsolete pre-#35 warnings, records current safe root validation semantics and remaining export mutation boundary, makes Storybook guidance workspace-aware, respects pinned Rust toolchains/supply-chain checks, and keeps #92 authoritative for exact live targets. No executable configuration or runtime code changes.

Issue #167, opened 2026-09-04
Migrate the virtual Rust workspace from resolver 2 to resolver 3 with lockfile evidence


xtreemze:
Parent: #145
Depends on: #147, #165 / #166
Related: #66

## Current state

The root virtual workspace explicitly uses:

```toml
resolver = "2"
```

while `defend-core` is edition 2024 and the project baseline is moving to Rust 1.98.1.

Rust's 2024 Edition Guide states that edition 2024 implies Cargo resolver 3 for package workspaces, which makes dependency resolution aware of `package.rust-version`. For a **virtual workspace**, however, Cargo cannot infer an edition from a root package, so resolver 3 must be selected explicitly in `[workspace]`.

## Why not change it in #166

#166 only makes the existing hybrid-lab standalone boundary valid via `workspace.exclude`. Resolver 2 -> 3 can select different dependency versions in the root workspace and therefore needs a real before/after dependency graph rather than being hidden in a structural one-line fix.

## Proposed migration

After the Rust 1.98.1 toolchain is certified and the current workspace boundary is valid:

1. generate/record the resolver-2 `defend-core` dependency graph with the actual Rust 1.98.1 Cargo;
2. create a narrow PR changing only root `resolver = "2"` -> `"3"` plus the real Cargo.lock delta it causes;
3. compare `cargo metadata` / `cargo tree` before and after;
4. specifically record any dependency whose selected version changes due to MSRV-aware fallback;
5. run fmt/clippy/test and `defend-core --features wasm --target wasm32-unknown-unknown`;
6. run supply-chain/security review against the resulting lockfile;
7. confirm the excluded hybrid runtime remains outside the root resolution graph.

## Dependency status checked 2026-09-04

No direct Rust-family update currently requires bundling into this migration:

- Bevy modular crates: 0.19.1 remains latest stable;
- wasm-bindgen: 0.2.127 remains latest;
- wasm-pack: 0.15.0 remains latest;
- Rapier JS: 0.20.0 remains latest.

Keep those version decisions separate so any resolver-induced lockfile delta is attributable to resolver semantics rather than direct dependency churn.

## Promotion rule

Adopt resolver 3 only if the dependency graph remains compatible with Rust 1.98.1 and the deterministic/WASM contract. If it changes a transitive version materially, preserve the before/after evidence on the PR rather than forcing resolver 3 merely because it is newer.

xtreemze:
[...]

PR #166, opened 2026-09-04
Make the hybrid Rust lab explicitly standalone

xtreemze:
Refs #165
Parent: #145
Related: #66, #67, #147

## Purpose

Make Cargo ownership match the architecture already intended by the hybrid-engine lab.

The root is a virtual Cargo workspace containing only `crates/defend-core`, while `experiments/hybrid-engine/rust` sits beneath the workspace root but owns its own package/release profile. Current Cargo validation explicitly errors when such a package discovers an ancestor workspace but is not a member, and recommends `workspace.exclude` or a nested `[workspace]`.

## Change

Add exactly:

```toml
exclude = ["experiments/hybrid-engine/rust"]
```

to the root `[workspace]`.

This makes the existing standalone intent explicit at the repository boundary.

## Why exclusion rather than membership

The hybrid runtime currently owns:

```toml
[profile.release]
codegen-units = 1
lto = true
opt-level = "s"
panic = "abort"
strip = "symbols"
```

Making it a root workspace member would move profile ownership to the root and change lockfile/target-directory/dependency-resolution semantics. #145 owns that later consolidation decision. This PR does not pre-commit to it.

A nested empty `[workspace]` in the hybrid manifest would also solve Cargo discovery, but root `exclude` is the clearer repository-wide declaration of intentional non-membership.

## Deliberately unchanged

- root workspace `members`;
- `resolver = "2"`;
- all Cargo package/dependency versions;
- hybrid release profile;
- Rust source;
- Cargo.lock state;
- #147 Rust 1.98.1 toolchain lane.

Resolver 2 -> 3 is deliberately separate. Rust 2024 recommends resolver 3 for virtual workspaces, but that changes dependency selection and should be measured with a real lockfile rather than bundled into this structural correction.

## Local certification required

Keep draft until the local executor records against this exact head:

1. `cargo metadata --manifest-path experiments/hybrid-engine/rust/Cargo.toml` succeeds rather than reporting workspace-membership mismatch;
2. the returned workspace root for that manifest is the hybrid crate itself / it is not a root-workspace member;
3. hybrid `cargo fmt --check`, clippy, test and wasm-pack build succeed under the pinned Rust 1.98.1 lane where practical;
4. root `cargo metadata` still lists only `defend-core`;
5. hybrid release-profile/build-output behavior is unchanged;
6. no Cargo.lock is committed as an incidental side effect.

## Promotion

[...]

Issue #165, opened 2026-09-04
Make the hybrid Rust lab explicitly standalone before Cargo consolidation


xtreemze:
Parent: #145
Related: #66, #67, #147

## Problem

The repository root is a virtual Cargo workspace with only:

```toml
[workspace]
members = ["crates/defend-core"]
resolver = "2"
```

`experiments/hybrid-engine/rust` is physically beneath that workspace root, has its own package manifest and release profile, but is neither a workspace member nor explicitly excluded.

Current Cargo workspace validation documents the corresponding failure mode as:

> current package believes it's in a workspace when it's not

and recommends either adding the package to `workspace.exclude` or adding a nested empty `[workspace]` table.

## Current architectural intent

The hybrid lab is deliberately standalone for now:

- it owns a local size/LTO/panic release profile used by wasm-pack;
- #145 has not yet decided how that profile should be represented if the crate becomes a root workspace member;
- absorbing it now would move profile ownership and dependency/lockfile semantics before local evidence exists.

Therefore the least-invasive correction is to make the existing intent explicit at the root:

```toml
exclude = ["experiments/hybrid-engine/rust"]
```

Do not add a nested `[workspace]` to the lab because the root exclusion is the more visible repository-wide ownership declaration.

## Scope

Narrow manifest-only PR:

- add the explicit root workspace exclusion;
- do not change workspace members;
- do not change resolver version;
- do not move release profiles;
- do not generate Cargo.lock;
- do not change Rust dependencies or source.

## Local gate

Against the exact PR head, under the pinned Rust 1.98.1 lane where practical:

1. `cargo metadata --manifest-path experiments/hybrid-engine/rust/Cargo.toml` succeeds and identifies the hybrid crate as standalone;
2. hybrid `cargo fmt --check`, clippy, test and wasm-pack build succeed;
3. root `cargo metadata` still contains only `defend-core`;
4. hybrid build artifacts/profile behavior remain local and unchanged;
5. no Cargo.lock is committed as a side effect unless a later dedicated lockfile PR owns it.

## Follow-up

After this boundary is certified, #145 may compare keeping the runtime standalone versus migrating it into the root workspace with explicit profile and lockfile ownership. Resolver 2 -> 3 is also a separate dependency-resolution decision; Rust 2024 recommends resolver 3 for modern virtual workspaces, but it should not be smuggled into this structural correction.

xtreemze:
## Local evidence confirming this issue, 2026-09-07

[...]

PR #168, opened 2026-09-04
Ignore Cargo target build artifacts repository-wide

xtreemze:
Refs #145
Related: #147, #165, #166, #167

## Purpose

Keep Rust build output from polluting repository/certification status while preserving lockfile visibility.

The root Cargo workspace writes compiled artifacts to the workspace-root `target/` directory. The repository-wide `.gitignore` did not ignore `target/`, and `crates/defend-core` has no local `.gitignore`. Ordinary root `cargo test`/`cargo check` can therefore create a large untracked tree during certification.

## Change

Add exactly:

```gitignore
target/
```

to the global build-output section.

Because the pattern is not anchored, it also safely covers nested Cargo `target/` directories. Existing hybrid-specific ignores can remain redundant without harm.

## Deliberately unchanged

- `Cargo.lock` remains visible/committable;
- `pnpm-lock.yaml` remains visible/committable;
- no source, manifest, dependency, build command, or runtime behavior changes.

## Validation

Static ignore-policy change only. Safe to merge after exact-head review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Build directory
 dist/
+target/
 
 # Logs
 logs


xtreemze (commented):
Exact-head static review PASS. Diff adds only repository-wide `target/` build-artifact ignoring; Cargo/pnpm lockfiles remain visible, and no source/manifest/command semantics change. Safe to merge without runtime certification.

PR #163, opened 2026-09-04
certification: combine modern pnpm workspace with safe Rust baseline

xtreemze:
Refs #92, #143, #145
Certification assembly for: #144 + #147 + #159
Related: #66, #149, #158, #162

## Purpose

Freeze one coherent **modernization bootstrap** for the infrequent local executor so the first shared pnpm lockfile, modern JS/browser checks, and Rust/WASM checks can be performed in one setup without ever invoking the known-bad Rust 1.98.0 compiler.

This is a certification/integration branch, not a new architecture lane. Source PRs remain provenance.

## Exact source cut

JS workspace parent:

- #144 head: `e367d1abad6018ccd884adda4220de4f5d37b49a`.

Rust safety source:

- #147 head: `91543cf8a340f95caaa4e01c46c7bf00d5b3f7c7`;
- #159 child head: `0e7f3399aed25ad97fdb3a8e70a90dcd46d71fd8`.

Assembly started at the exact #144 head. The #147/#159 Rust delta was then copied without changing its semantics.

Compared directly with #144, this branch adds only:

- `rust-version = "1.98.1"` to `defend-core`;
- hybrid runtime `rust-version` 1.98 → 1.98.1;
- repository-root `rust-toolchain.toml` with minimal 1.98.1 + clippy/rustfmt + `wasm32-unknown-unknown`;
- deletion/move of the nearer hybrid `rust-toolchain.toml` that was pinned to 1.98.0.

GitHub compare reports four commits ahead / zero behind #144 and exactly those Rust toolchain/MSRV changes.

## Why this batch is safer than certifying #144 and #147 independently

The #144 hybrid build invokes wasm-pack from the hybrid package. On the raw #144 head, rustup discovers the nested hybrid `rust-toolchain.toml` first, selecting Rust 1.98.0. That would make a successful workspace/WASM test unsafe evidence.

This assembly gives the local executor one checkout where:

- modern pnpm workspace ownership is fixed;
- the hybrid and root Rust commands share one 1.98.1 authority;
- the unsafe nested override is absent;
- no Cargo workspace/profile migration has yet occurred;
- the historical root dependency graph remains untouched.

## Local phase 1 — establish tool identities

Record exact SHA, OS/arch and clean status, then:

```sh
node --version
pnpm --version
wasm-pack --version
rustup show active-toolchain
rustc -Vv
cargo -V
```

Repeat the Rust identity commands from:

```sh
cd experiments/hybrid-engine/rust
rustup show active-toolchain
rustc -Vv
[...]

Issue #158, opened 2026-09-04
Evaluate pnpm 12 after the modern workspace baseline is certified


xtreemze:
Parent: #66
Depends on: #143 / #144
Related: #154, #155

## Why not upgrade during workspace bootstrap

pnpm 12 is now released, but the npm package still advertises 11.25.0 on the default `latest` tag while pnpm 12 is carried on its major-specific channel. The 12.0 release is also a new Rust/pacquet implementation with meaningful behavior changes, and post-release compatibility issues have already been accepted upstream (including formerly supported `pnpm install` flags being rejected).

#144's purpose is to establish the first reproducible shared lockfile and workspace ownership. Changing package-manager major at the same time would make any resolution/configuration discrepancy ambiguous.

## Current decision

Keep the certified baseline on:

- Node 24 LTS;
- pnpm 11.25.0;
- one `experiments/pnpm-lock.yaml` generated by that exact package-manager version.

Do not interpret this as a long-term rejection of pnpm 12.

## pnpm 12 evaluation after #144

Create a dedicated branch from a certified #144/master state and compare pnpm 11.25.0 with the then-current pnpm 12 patch release.

Required checks:

1. shared-workspace lockfile importers/resolution remain complete for both modern packages;
2. `saveExact`, `sharedWorkspaceLockfile`, `engineStrict`, and release-age policy are recognized and enforced;
3. `pnpm install --frozen-lockfile` rejects stale member manifests correctly;
4. recursive/filter commands used by aggregate scripts behave identically;
5. Storybook Playwright/browser project resolution remains healthy;
6. hybrid-engine WASM script/filter execution remains healthy;
7. no unexpected lifecycle/build-script execution appears;
8. compare install time, store size, lockfile churn/format, and diagnostics;
9. confirm Dependabot/automation can update the new lockfile format;
10. review upstream accepted pnpm 12 regressions relevant to our command surface before promotion.

## Promotion rule

Only change the workspace `packageManager` after the pnpm 11 lockfile baseline exists and the same dependency graph/tests are demonstrated under pnpm 12. Keep the migration isolated from Babylon, Storybook, physics, and gameplay dependency updates.

xtreemze:
[...]

Issue #66, opened 2026-09-03
Modernize Defend around a Babylon 9 renderer and modular Bevy/Rust simulation workspace


xtreemze:
Parent: #29
Related: #31, #52, #53, #56, #60, #64

## Direction

Modernize the project as a **hybrid web/native engine workspace**, not as a monolithic full-engine rewrite.

The strongest current architecture is:

- **Babylon.js 9.x** owns browser rendering, picking, camera, WebGL/WebGPU, presentation diagnostics, and accessibility integration.
- **Rust/WASM** owns deterministic gameplay/simulation seams where it materially improves correctness, testing, portability, or performance.
- Use **modular Bevy crates** (`bevy_ecs`, `bevy_app`, `bevy_time`, `bevy_math`, optionally `bevy_transform`) rather than bringing Bevy's renderer/window stack into the browser path by default.
- Evaluate **Rapier 0.20** as the primary cross-web/native physics candidate, with deterministic builds used for replay/certification where the cost is acceptable.
- Keep the object-based audio contract independent from both engines; AudioWorklet/Web Audio remains the browser renderer, with optional Rust/WASM DSP.
- A full Bevy renderer remains an alternate native/web frontend experiment, not something that should render the same production browser scene alongside Babylon.

This gives us Bevy's ECS/scheduling/time model and Rust portability without paying for two 3D renderers or duplicating GPU state.

### Browser / tooling
- Node 24.20.0 LTS (Node 26.8.1 is Current; prefer LTS for project baseline)
- pnpm 11.25.0
- Vite 8.2.2 (Rolldown-based)
- TypeScript 7.0.2
- Biome 2.5.11
- Vitest 4.1.11
- Playwright 1.62.1
- Storybook 10.6.x

### Babylon
- `@babylonjs/core` 9.23.0
- `@babylonjs/inspector` 9.23.0
- `@babylonjs/accessibility` 9.23.0
- optional `@babylonjs/havok` 1.3.14 for comparison fixtures

Use ESM packages and tree-shakeable imports. Babylon explicitly recommends `@babylonjs/core` for new projects; 9.x supports WebGL and WebGPU side-by-side. Babylon 9 also exposes headless Inspector/CLI tooling suitable for AI agents.

### Rust / Bevy / WASM
- Rust 1.98.0 stable, edition 2024
- Bevy 0.19.1 family
- `bevy_ecs` 0.19.1
- `bevy_app` 0.19.1
- `bevy_time` 0.19.1
[...]

PR #169, opened 2026-09-04
build: consolidate Rust workspace without changing dependency versions

xtreemze:
Refs #145
Parent: #163
Related: #66, #143, #144, #149, #158

## Purpose

Consolidate the modern Rust crates into one root Cargo workspace while preserving the hybrid WASM release semantics exactly and leaving dependency versions unchanged.

This PR is intentionally stacked on #163. It must not be retargeted to `master` until the bootstrap/toolchain baseline is certified and promoted.

## Changes

- add `experiments/hybrid-engine/rust` as a root workspace member;
- keep `crates/defend-core` as the default member so ordinary root Cargo commands retain their current scope;
- retain resolver `2` for this migration;
- move the hybrid size/LTO/panic/strip settings into a dedicated root `hybrid-wasm` profile;
- remove the hybrid crate's now-non-authoritative local `[profile.release]`;
- update the hybrid wasm-pack command to use `--profile hybrid-wasm`;
- no Bevy, wasm-bindgen, Rapier, Babylon, or other dependency-version changes;
- no `Cargo.lock` is synthesized in GitHub. The real lockfile must be generated and inspected with Cargo 1.98.1 during local certification.

## Why this shape

Cargo reads profile settings from the workspace root once the hybrid crate becomes a member. A per-package release-profile override cannot preserve `lto` and `panic`, so broadening root `[profile.release]` would silently impose the hybrid's size-oriented settings on every current/future crate. A dedicated custom profile keeps those semantics scoped to the WASM build.

`default-members = ["crates/defend-core"]` prevents a plain root `cargo test/check/build` from unexpectedly expanding from one crate to every workspace member.

## Local certification gate

After #163 is certified/promoted, test this exact head with Rust/Cargo 1.98.1:

1. `cargo metadata --no-deps` confirms both members and `defend-core` default membership;
2. generate the first real root `Cargo.lock`; inspect it before commit;
3. run `cargo tree --workspace` and `cargo tree -d`, with explicit attention to wasm-bindgen resolution;
4. run root fmt/clippy/test/check for `defend-core`;
5. run fmt/clippy/test for `defend-hybrid-runtime` through the root workspace;
6. build the hybrid with the updated wasm-pack command;
7. compare `.wasm` bytes/compressed size against the pre-consolidation standalone release build;
8. exercise the hybrid browser fixture and Inspector path;
9. verify no Bevy renderer/window stack appears in the dependency tree;
10. verify the historical npm/Webpack graph remains untouched.

Do not centralize `[workspace.dependencies]` or change dependency ranges in this PR. That should follow only after the generated lockfile/tree establishes the actual resolved graph.


xtreemze:
Retargeted to current `master` now that #163 is superseded. The historical intended delta was exactly three files (`Cargo.toml`, hybrid `package.json`, hybrid `rust/Cargo.toml`), but it cannot be replayed verbatim anymore: current `master` has moved the root workspace to Cargo resolver 3, has a committed root `Cargo.lock`, and the hybrid crate now carries `[package.metadata.wasm-pack.profile.release]` flags required by the current wasm-pack/Binaryen path. Reassembly must therefore preserve resolver 3 unless there is a measured reason to regress it, preserve/translate the wasm-pack compatibility flags for the custom profile, and regenerate/inspect the real root lockfile with both workspace members. Do not merge the current 11-file retargeted stale diff. Keep draft until a clean current-master migration head and Rust/WASM/browser evidence exist.

xtreemze:
2026-09-13 reassembly checkpoint: keep this PR draft and do not merge the current stale diff. Current `master` is resolver 3 with a committed root Cargo.lock and a standalone hybrid release profile plus wasm-pack metadata. A second prerequisite is now explicit: hosted CI still used wasm-pack 0.13.1, while this PR's intended `--profile hybrid-wasm` build requires custom-profile support introduced after that baseline. PR #201 updates the hosted/documented baseline to wasm-pack 0.15.0 without changing workspace semantics. After #201 passes/promotes, rebuild this consolidation from current `master`: retain resolver 3 and `default-members = ["crates/defend-core"]`; add the hybrid runtime as a workspace member; move the size/LTO/panic/strip settings to a root custom profile; translate/preserve the current wasm-opt feature metadata for that custom profile; regenerate and inspect the real root Cargo.lock; remove the now-obsolete nested hybrid lockfile only if the consolidated Cargo run proves root ownership; then attach fmt/clippy/test/WASM/browser evidence. Until then this branch remains provenance, not an integration candidate.

Issue #29, opened 2026-09-03
Modernization program: preserve Defend's physics-driven resource loop while evolving the game


xtreemze:
## Purpose

Use this issue as the architectural/design coordination thread for modernizing Defend. The implementation is old, but the game's distinctive mechanics should be treated as behavioral requirements rather than incidental legacy code.

## Core gameplay contract

Defend is a physics-driven resource-defense game, not a conventional lane/path tower-defense game.

The following principles should remain true unless we explicitly decide otherwise here:

- **Energy is simultaneously health, construction budget, and the resource being defended.**
- **Offense sustains defense.** Projectile hits recover energy, so active combat is part of the economy rather than only damage output.
- **Defenses are temporary.** Towers age; higher levels eventually degrade to lower levels and disappear.
- **Physics is gameplay.** Mass, inertia, collision, obstruction, projectile impulse, knockback, and pushing enemies off the platform must remain strategically meaningful.
- **Enemies are finite-lived agents.** Their health decays over time, so delay/redirection can be valid strategies alongside damage.
- **Threat mitigation is plural.** A useful action may destroy, weaken, delay, block, redirect, or eject an enemy.
- **Higher tower levels should represent different physical/temporal trade-offs, not merely larger numbers.**
- **Direct world interaction is preferred.** Placement/upgrading should remain grounded in manipulating the battlefield rather than becoming menu-heavy.
- **The aesthetic should remain procedural/abstract where practical.** Generated geometry, materials, effects, and spatial/procedural audio are part of the identity.
- **Performance is a gameplay constraint.** The simulation should remain viable on constrained/mobile hardware and degrade effects before compromising core mechanics.

## Working model for parallel AI executors

We expect multiple highly capable agents to work simultaneously.

- GitHub **issues** are the durable place for design decisions, investigations, disagreements, findings, and future work.
- GitHub **PRs** are the unit of implementation. Prefer small, independently reviewable PRs with narrow ownership.
- Before starting a PR, inspect open issues/PRs for overlap and state the intended ownership boundary.
- Do not couple unrelated modernization work into one branch merely because the old code is entangled.
- When a change exposes a nearby defect that is safe and directly related, fixing it in the same PR is acceptable; otherwise open/follow an issue.
- Avoid speculative compatibility shims for obsolete dependencies when a deliberate replacement is planned.

[...]

PR #182, opened 2026-09-07
Make 'pnpm check' reproducible in the hybrid lab and fix its diagnostics

### LINKED ISSUE #174: Make 'pnpm check' reproducible in the hybrid lab and fix its 20 real biome diagnostics
Found during the 2026-09-07 local certification campaign on #163 head `c1318a946a841faeb90e7ab6500fb8a3ee20b276`. Full evidence: https://github.com/xtreemze/defend/pull/163#issuecomment-5563899938

Two related defects in the same file. Both make `pnpm --filter @defend/hybrid-engine-lab check` unusable as a gate.

## Defect 1 — `check` is not reproducible across a Rust build

`experiments/hybrid-engine/biome.json` declares:

```json
"files": { "includes": ["**", "!pkg", "!dist", "!node_modules"] }
```

`rust/target` is not excluded. Measured on the same source tree, same command, same biome 2.5.11:

| State | Files checked | Result |
|---|---|---|
| clean checkout | 13 | 20 errors, 4 warnings |
| after any `cargo`/`wasm-pack` build | 288 | 289 errors |

The extra 269 come entirely from `rust/target/**/*.json` build fingerprints. So `pnpm check` reports a different answer depending on whether a Rust build has run — which it always will, since `pnpm build` runs `pnpm wasm` first.

Fix: add `!rust/target` to `files.includes`, or enable `vcs.useIgnoreFile` so biome honours the repository `.gitignore` (which already ignores `target/`, since #168).

## Defect 2 — 20 real lint errors on a clean tree

With `rust/target` excluded, the genuine result is **20 errors, 4 warnings, 1 info** across 13 files:

| Rule | Count |
|---|---|
| `lint/suspicious/useIterableCallbackReturn` | 8 errors |
| `assist/source/organizeImports` | 3 errors |
| `lint/a11y/useAriaPropsSupportedByRole` | 1 error |
| `lint/style/useImportType` | 3 warnings |
| `lint/complexity/useOptionalChain` | 1 warning |

Files: `mothership.html`, `src/fixedStep.test.ts`, `src/geothermalPrototype.ts`, `src/main.ts`, `src/mothershipNavigationPrototype.ts`, `src/surfaceRoutingPrototype.ts`, `src/towerTerrainPrototype.ts`.

`useIterableCallbackReturn` is the interesting one — it is a correctness rule, not style. Example at `src/main.ts:464`:

```ts
towerPositions.forEach((position) => stabilizeFoundation(position));
```

the arrow body returns `stabilizeFoundation`'s value into a `forEach` that discards it. Each of the 8 sites should be inspected rather than blanket-autofixed: some may be a genuinely misused `forEach`/`map`.

## Suggested boundary

Two commits, or two PRs if the lint fixes turn out to touch prototype behavior:

1. the `biome.json` exclusion (mechanical, no source change);
2. the 20 diagnostics, each reviewed.

Do not fold either into #163.

## Acceptance

- `pnpm --filter @defend/hybrid-engine-lab check` returns the same result before and after a Rust build;
- it exits 0 on a clean checkout;
- no `forEach`/`map` site was "fixed" by discarding a return value that was actually meaningful.

xtreemze:
## Scope

Linked issue: closes #174

Ownership boundary: `experiments/hybrid-engine/biome.json`, `mothership.html`, and the lab's `src/*.ts`. No dependency, build-config, or gameplay changes.

**Stacked on #179** (which is stacked on #178), because #179 edits the import blocks in the same six source files and `organizeImports` here would conflict head-on otherwise. The diff against that parent is only this PR's three commits.

Three commits, deliberately separable:

| Commit | What |
| --- | --- |
| `5703b81` | `biome.json` — stop checking Rust build artifacts, drop a deprecated field |
| `09f7714` | the 12 real lint/assist errors |
| `607ecb0` | the formatting sweep, mechanical |

### 1. `pnpm check` was not reproducible

| State | Files checked | Result |
| --- | --- | --- |
| clean checkout | 13 | 20 errors |
| after any Rust build | 288 | 289 errors |

The extra 269 were `rust/target/**/*.json` fingerprints. Since `pnpm build` runs `pnpm wasm` first, the gate effectively never saw the clean-tree answer.

`biome.json` excluded `pkg`, `dist` and `node_modules` by hand but not `rust/target`. Rather than adding a fourth entry to a list that has already drifted once, this enables `vcs.useIgnoreFile`: the package's `.gitignore` already lists all four, so there is now one source of truth instead of two that must be kept in sync. Verified with `rust/target` present on disk: 19 files checked, result identical to the clean tree.

Also migrates `linter.rules.recommended` to the `preset` field it was deprecated in favour of, after confirming the enabled rule set is unchanged.

### 2. The 12 real errors

**`useIterableCallbackReturn` (8).** Every site is `array.forEach((x) => sideEffect(x))`, where the concise arrow body implicitly returns the callee's value into a `forEach` that discards it. As the issue asked, each was inspected rather than blanket-autofixed — all eight callees return `void` (`Mesh.dispose`, `updateRaider`, `updateEnergy`, `stabilizeFoundation`), so none was a `forEach` that should have been a `map`. Block bodies, behaviour-preserving.

**`useAriaPropsSupportedByRole` (1).** `<div id="controls">` carried an `aria-label` with no role to support it. `role="toolbar"` is the accurate role for a row of control buttons and does support a label, so the author's intent is kept rather than dropping the label. `role="group"` was tried first and traded the error for `useSemanticElements`, which wants `<fieldset>` — the wrong element for a toolbar containing no form controls.

**`organizeImports` (3)** and **`useImportType` (3 warnings)** applied via biome's safe autofixes; three `Mesh` imports are used only as types.

**Deliberately not applied:** `useOptionalChain` at `geothermalPrototype.ts:703`. Biome marks that fix unsafe, and rewriting `!source || source.phase !== "active"` as an optional chain is a readability judgement on a guard clause, not a defect. It is a warning, does not fail the gate, and is left for whoever owns that prototype.

### 3. The formatting sweep

`biome check` includes formatting and 8 of 13 files had never been run through it, so the gate could not pass on formatting alone. This is the large commit — 825 insertions — and it is entirely mechanical.

I considered raising `formatter.lineWidth` to shrink the diff and rejected it: the sources are not written to any consistent width (p50 28, p90 70, p99 106, **max 272** columns), so there is no author intent to preserve, and `biome.json` already accepted every formatter default except `indentStyle`.

## Gameplay impact

- [x] No intended gameplay/balance change

No #29 invariant is touched. The eight `forEach` conversions were each verified against a `void`-returning callee, and all six lab pages were re-run in a browser afterwards.

## Validation

macOS 26.5.2 aarch64, Node 24.20.0, biome 2.5.11, Chromium 151.0.7922.34.

- `biome check .` → **exit 0**, 0 errors (was 20), 1 intentional warning;
- result is now identical with and without `rust/target` on disk;
- `tsc --noEmit` → clean;
- `vitest run` → 5 passed;
- `pnpm build` → green;
- six-page headless smoke, zero console errors and a progressing render loop on every page, `index.html` at 31.8 fps — run after the lint commit **and** again after the formatting commit.

Not validated online / requires local Codex:

- [x] build/install — done locally
- [x] browser interaction — done locally, all six pages
- [ ] physics behavior
- [ ] performance/profile
- [ ] touch/input
- [ ] audio
- [ ] PWA/offline

## Concurrency

Known overlapping PRs/issues: #178 and #179 (this stack), #163 (its `pnpm check` gate fails until this lands).

Integration notes for other executors:

- **Conflict exposure is low but time-sensitive.** I checked every open PR: only #179, this PR's own parent, touches `experiments/hybrid-engine/src/`. The formatting commit will conflict with anything new that lands in those files, so landing it sooner is cheaper than later. It is isolated in its own commit if it needs to be dropped or re-run.
- No lockfile is committed. Verification used a package-local `pnpm install`; the generated lockfile was left untracked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/experiments/hybrid-engine/biome.json
+++ b/experiments/hybrid-engine/biome.json
@@ -1,7 +1,12 @@
 {
   "$schema": "https://biomejs.dev/schemas/2.5.11/schema.json",
+  "vcs": {
+    "enabled": true,
+    "clientKind": "git",
+    "useIgnoreFile": true
+  },
   "files": {
-    "includes": ["**", "!pkg", "!dist", "!node_modules"]
+    "includes": ["**"]
   },
   "formatter": {
     "enabled": true,
@@ -10,7 +15,7 @@
   "linter": {
     "enabled": true,
     "rules": {
-      "recommended": true
+      "preset": "recommended"
     }
   }
 }
--- a/experiments/hybrid-engine/mothership.html
+++ b/experiments/hybrid-engine/mothership.html
@@ -61,7 +61,7 @@
   <body>
     <div id="app"><canvas id="renderCanvas"></canvas></div>
     <div id="metrics">initializing…</div>
-    <div id="controls" aria-label="Mothership prototype controls">
+    <div id="controls" role="toolbar" aria-label="Mothership prototype controls">
       <button id="reset" type="button">Reset [R]</button>
       <button id="pause" type="button">Pause drain [Space]</button>
       <button id="fast" type="button">Fast drain [F]</button>
--- a/experiments/hybrid-engine/src/fixedStep.test.ts
+++ b/experiments/hybrid-engine/src/fixedStep.test.ts
@@ -12,7 +12,10 @@ const policy: FixedStepPolicy = {
   maxFrameDeltaSeconds: 0.25,
 };
 
-function simulateFrames(frameCount: number, frameDeltaSeconds: number): {
+function simulateFrames(
+  frameCount: number,
+  frameDeltaSeconds: number,
+): {
   ticks: number;
   state: FixedStepState;
 } {
@@ -76,7 +79,10 @@ describe("advanceFixedStep", () => {
 
   it("sanitizes non-finite timing inputs", () => {
     const advance = advanceFixedStep(
-      { accumulatorSeconds: Number.NaN, droppedSeconds: Number.POSITIVE_INFINITY },
+      {
+        accumulatorSeconds: Number.NaN,
+        droppedSeconds: Number.POSITIVE_INFINITY,
+      },
       Number.NaN,
       {
         fixedDeltaSeconds: Number.NaN,
--- a/experiments/hybrid-engine/src/fixedStep.test.ts
+++ b/experiments/hybrid-engine/src/fixedStep.test.ts
@@ -1,9 +1,9 @@
 import { describe, expect, it } from "vitest";
 import {
-  INITIAL_FIXED_STEP_STATE,
   advanceFixedStep,
   type FixedStepPolicy,
   type FixedStepState,
+  INITIAL_FIXED_STEP_STATE,
 } from "./fixedStep";
 
 const policy: FixedStepPolicy = {
--- a/experiments/hybrid-engine/src/geothermalPrototype.ts
+++ b/experiments/hybrid-engine/src/geothermalPrototype.ts
@@ -323,12 +323,23 @@ function createTower(
 
 function conduitPath(source: GeothermalSource, tower: TowerProxy): Vector3[] {
   const start = source.position.add(new Vector3(0, 1.2, 0));
-  const end = tower.mesh.position.add(new Vector3(0, -tower.mesh.getBoundingInfo().boundingBox.extendSizeWorld.y + 0.7, 0));
+  const end = tower.mesh.position.add(
+    new Vector3(
+      0,
+      -tower.mesh.getBoundingInfo().boundingBox.extendSizeWorld.y + 0.7,
+      0,
+    ),
+  );
   const lateral = end.subtract(start);
   const side = new Vector3(-lateral.z, 0, lateral.x);
-  if (side.lengthSquared() > 0.001) side.normalize().scaleInPlace(Math.min(5, lateral.length() * 0.12));
-  const p1 = Vector3.Lerp(start, end, 0.32).add(side).add(new Vector3(0, 1.4, 0));
-  const p2 = Vector3.Lerp(start, end, 0.68).subtract(side.scale(0.45)).add(new Vector3(0, 1.1, 0));
+  if (side.lengthSquared() > 0.001)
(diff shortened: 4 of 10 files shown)


xtreemze:
### Rebased onto `master` after #179 was squash-merged — re-certified at `c0e3e56`

#178 and #179 landed while this was open, and the squash merge left this branch carrying a commit whose content was already on `master` under a different SHA, so GitHub reported `CONFLICTING` / `DIRTY`.

Resolved by replaying only this PR's own three commits onto current `master` `4b59573`:

```
git rebase --onto origin/master e45cf61 agent/hybrid-biome-scope
```

No conflicts. The dropped commit is #179's, now on `master` as `4b59573`. Pushed with `--force-with-lease` against the exact prior head `607ecb0`; the force update is on a branch this executor owns and its reason is the squash-merged parent, not a behind count. Base is now `master`, so this is no longer stacked.

The three commits are unchanged in content and the diff against `master` is identical to what was reviewed — same 10 files, same +863/−216.

| Commit | |
| --- | --- |
| `d01971e` | `biome.json` — stop checking Rust build artifacts, drop a deprecated field |
| `3195ee0` | the 12 real lint/assist errors |
| `c0e3e56` | the formatting sweep, mechanical |

### Re-verified at the new head

macOS 26.5.2 aarch64, Node 24.20.0, biome 2.5.11, rustc 1.98.1, Chromium 151.0.7922.34.

- `biome check .` → **exit 0** with `rust/target` present on disk, 19 files, 0 errors, 1 intentional warning;
- `tsc --noEmit` → clean;
- `vitest run` → 5 passed;
- `pnpm build` → green;
- six-page headless smoke → **6 of 6 PASS**, zero console errors, progressing render loop on every page, `index.html` at 29.7 fps.

Evidence applies to `c0e3e56` specifically, per §7.

### Conflict exposure

Unchanged and still low: no other open PR touches `experiments/hybrid-engine/src/`. The formatting commit is isolated, so it can be dropped and re-run if something lands there first.

Issue #106, opened 2026-09-04
Build causal-legibility fixtures and prediction/comprehension evidence


xtreemze:
Parent: #103
Related: #30, #72, #76, #86, #92, #105

## Purpose

Turn the causal-legibility contract from #103 into executable design evidence without changing production balance.

The key question is not merely whether the simulation is deterministic. It is whether a player can increasingly **predict**, **observe**, and **explain** important physical/economic outcomes from world cues.

This belongs to a later certification campaign than the frozen #92 cut unless explicitly run as an independent lab lane.

## Design hypothesis

A strategically important interaction should expose three readable phases:

`anticipation → causation → residue`

Examples:

- T3 winds/aims/commits → heavy projectile transfers impulse → altered trajectory/ejection remains obvious;
- R3 carries visible inertia → cannot negotiate an R1 opening → collision/route consequence persists;
- tower approaches degradation → degrades → changed firing/geometry state remains readable;
- energy spill approaches a blocked route → pools against geometry → stranded value remains visible;
- mothership reserve diminishes → lift/handling degrades → crash/hulk remains as residue.

The player should not need numeric debug UI to understand these chains, though development instrumentation may expose exact values.

## Recommended experimental surface

Prefer deterministic Storybook/hybrid-engine playgrounds before production integration.

Build a small set of canonical fixtures with repeat/reset controls and optional developer overlays:

1. **Impulse packet comparison** — T2 frequent-small vs T3 infrequent-heavy projectile events against the same body.
2. **Mass/inertia comparison** — same impulse against R1/R2/R3.
3. **Corridor fit** — R1-only, R1/R2, and broad R3 route geometries.
4. **Delay-as-damage** — barrier detour with viability/lifetime decay visible in-world.
5. **Drainage consequence** — open, diverted, sealed, and released teal-energy flow.
6. **Lifecycle consequence** — tower warning → degradation → disappearance with battlefield effect.
7. **Commitment residue** — failed expensive raider/mothership outcome leaving persistent physical evidence.

Do not turn these into tutorial popups. They are laboratories for finding the minimum world cues needed for understanding.

## Instrumentation

Each fixture should expose machine-readable snapshots/events sufficient to measure:

- initial state;
- predicted qualitative outcome category;
- actual outcome category;
- displacement/vector change;
- route selected / obstacle contacts;
- time-to-silo or time-to-expiry;
- energy released / collected / stranded / lost;
- causal event sequence timestamps;
- final state/residue.

[...]

PR #203, opened 2026-09-13
Add causal-legibility evidence contract and fixture matrix

xtreemze:
Refs #106
Related: #72, #86, #103, #113, #137

## Purpose

Give the causal-legibility design lane an executable evidence contract without creating another simulation authority.

This PR evaluates **observed, timestamped output from authoritative gameplay** as a semantic chain:

`anticipation → causation → residue`

It asks whether the cause of a consequential result was visible/readable before and during the event and whether an observable consequence remained afterward. It does **not** decide whether physics, combat, economy, ejection, expiry, or energy transfer were correct.

## Pure contract

Adds `src/js/gameplay/causalLegibility.ts` with:

- timestamped evidence events keyed by one semantic cause;
- explicit `anticipation`, `causation`, and `residue` phases;
- deterministic ordering independent of input-array order;
- fail-closed malformed evidence handling;
- distinct-channel counting so a fixture can require redundant presentation rather than color-only evidence;
- configurable anticipation-lead and residue-delay windows;
- per-chain diagnostics for missing phases, ordering/timing failures, and insufficient presentation channels;
- no Babylon, Bevy, physics, economy, timer, Worker, AudioContext, or production mutation.

## Storybook matrix

Adds `Foundations/Strategy/Causal Legibility` covering #106's seven canonical experimental families:

1. projectile packet comparison;
2. raider mass/inertia comparison;
3. R1/R2/R3 corridor fit;
4. delay into finite-life expiry;
5. teal drainage vs blocked pooling;
6. tower lifecycle consequence;
7. expensive commitment residue/capital risk.

It also adds two cross-system extensions already required elsewhere in the design backlog—edge ejection and fortress/world residue—and a tenth intentionally broken control with causation/residue but no readable anticipation.

The examples are semantic witness data, not claimed physics measurements. Later #72/#86 live fixtures should emit measured events into the same contract rather than copying these example numbers into gameplay.

## Scope

Current branch is two added files only, based directly on current `master`:

- `src/js/gameplay/causalLegibility.ts`;
- `experiments/storybook/src/Foundations/Strategy/CausalLegibility.stories.ts`.

No dependencies, production call sites, balance constants, existing certification heads, or authoritative simulation contracts change.

## Validation gate

Keep draft until the current root/Storybook toolchain verifies:

```sh
npm run lint
npm run build
cd experiments/storybook
pnpm typecheck
pnpm build
pnpm test
```

Also verify:

- reordering the same semantic event set does not alter the result;
- malformed/non-finite evidence fails closed;
- the nine positive semantic witness rows qualify under the default evidence policy;
- the deliberately opaque control does not qualify;
- stricter channel/timing controls visibly expose why a chain loses qualification.

## Integration boundary

This is an evidence/presentation seam only. #106 remains open after this PR: completing it still requires live measured outputs from #72/#86 and human/browser review proving players can infer causation without debug overlays.


xtreemze:
Hosted check update for head `5111425c869fa646e01baa845794d319d4c6b750`:

- GitHub Pages / legacy production build: **PASS** (run #6).
- This is not sufficient to promote the PR: the current Pages workflow runs `npm run build` only and does not typecheck/build/test `experiments/storybook`.

Keep draft until the Storybook package gate in the PR body is executed. The design evidence contract is intentionally not being certified by an unrelated green production build.

